PR #133, opened 2022-11-22
Add icon to combo box

### LINKED ISSUE #130: (OUI Component Refresh) Add prop to set OuiIcon type=search true or false

BSFishy:
### Description
Add a search icon prop to the combo box

2022-11-22 10_15_22-Combo box - OpenSearch UI Framework
 
### Issues Resolved
#130 
 
### Check List
- [x] All tests pass
  - [x] `yarn lint`
  - [x] `yarn test-unit`


ashwin-pc:
I've called this out in the parent issue too, but i think we should not limit this to just the search icon, but rather any Oui Icon. even custom ones.

KrooshalUX:
@ashwin-pc can you provide a "why" on not limiting the icon? For an icon in the let placement in a search bar, I would find it confusing to be any other icon. I am providing this guardrail to prevent a whacky user experience from being implemented. Would like to hear your perspective though!

ashwin-pc:
Its just my opinion, but i want the design system to offer me as much flexibility as possible while letting me do the right thing easily. By changing the property to simply `icon`, It lets me do both. When I want to indicate search, I can use `icon="search"`, but is i want a combobox that indicates a list of tags that I am selecting, id prefer the `icon="tag"` option since thats more appropriate there. This lets me add the search icon when needed without making the flag seem like an afterthought having no other purpose than to just add a single icon to the input.

If this change was perhaps tied to a search feature in the combobox, i'd be more okay with the limitation since adding a prop like `search=true` indicates that this combobox will now act as a multi select search input and not just show an icon.

KrooshalUX:
@ashwin-pc - These are some good points about implementation, while I disagree about your tag use case, I do see a case where no tags pre-exist and therefore search is invalid, and add / plus is more appropriate. I know this is getting into splitting some hairs. After some re-thinking and discussion, here are my updated thoughts to the approach of having the icon prop. I want engineering maintainers within the project to set the path for implementation and developer experience - so these are for consideration given the feedback in the comments. 

Approach 1: 
- Add the icon prop type as configurable but to search by default
- Allow icon display to true/false
This approach gives UX the opportunity to set a "common sense default" without having to be heavy handed in implementation approach. It does add some friction but preserves the freedom of choice, whether I agree with it or not ;)  

Approach 2:
- Keep as-is and mark the changes with EXPERIMENTAL in order to measure impact on developer experience and UX. There is no reason that OUI cannot have the same approach to component changes as the OpenSearch application has. EUI even had a "beta" for their component changes as well, signaling that some prop definitions were not yet set in stone.

ashwin-pc:
I like both approaches but i'd lean towards option 1.1 where we add an icon prop. But I dont think adding search by default is a good idea. It would mean that to disable an icon in a combobox, we would have to explicitly pass `icon={false}` to not show one. Not very fond of this as a pattern. 

One other option would be to use the prop to define the purpose of the UI element. e.g. `<OuiComboBox as="search" />` This styles the component as a search only component that cannot add new values dynamically. With this prop on, a search icon being always present makes sense. And if we want to add additional flexibility, we can add an icon prop too that lets the user override this icon or disable it all together.

KrooshalUX:
For clarity & my education (thanks for taking me on the journey) - we wouldn’t want two separate props 1) icon type and 2) show/hide icon 

Vs  icon = false? 

@bsfishy any approach ideas here as well?

kgcreative:
I would be supportive of:
* Keeping the default API the same
* Adding an `icon` property
* For ease of use, `icon`, `icon=true` and `icon="search"` should be equivalent <- (I think this is what @KrooshalUX meant when she said "search" as default -- not search icon on by default on the component, but rather, when the icon prop is used, we should show the search icon if no other information is passed)
* `icon="false"`, or not even passing the `icon prop` should result in the component not rendering it.

ashwin-pc:
@kgcreative I can get behind that. Still not 100% onboard with the need to have `icon=true` default to search, but it does not take away from whats expected of the `icon` property.

kgcreative:
I think one conversation to have is which icon makes the most sense to be "default" icon. When I look at the list of glyphs, the ones that come to mind are `tag`, `dot`, `email` `apps` `home` and `search`, and looking at the usage of combobox in the application, `search` seems to be the most common case. I would rather `icon=true` not return an error, so more than anything this is graceful degradation

KrooshalUX:
@kgcreative can you provide cases for setting those other options as default - my main justification for setting the default to search it is the primary interaction within the input box - search to find a tag (object,data source,index, etc) or to search & receive no result in the case where custom tag is added. To match the current default of the input box and to not force a braking visual change the icon would not be visible by default. This allows us to determine the “correct” approach across various projects.

AMoo-Miki:
I am having trouble imagining a form that has 5 combo-boxes, all with a search icon to their left.
I wouldn't mind if the icon appeared only when active but doesn't an icon inside the field denote that that field is to search for something? Am i too old?

kgcreative:
> I am having trouble imagining a form that has 5 combo-boxes, all with a search icon to their left.
> 
> I wouldn't mind if the icon appeared only when active but doesn't an icon inside the field denote that that field is to search for something? Am i too old? 

The intent is that a combo box with an icon be more an exception that the rule, but when adding an icon, often time the search icon is the most appropriate (for example for searching/ multi selecting an index, a field, or a tag)

ashwin-pc:
@kgcreative @KrooshalUX The primary concern i have with this is the API. having a boolean prop that simply shows a search icon is indicative of code smell. If as @AMoo-Miki  said, the icon showed up during user input or when it active automatically, thats fine. Or if the prop allowed for arbitrary icons too, it seems like a good use of a prop, but simply using it to show or hide a single icon "search" seems wasteful and indicative that we havent abstracted the problem well. 

If we are doing this to not introduce a breaking change, thats something we should avoid since technically we can always find a solution does not break semver if we bend over backwards enough. But we should weigh that with the consequence of doing so. In this case its an API that serves a very specific usecase that imo can be solved without it.

API changes are hard to walk back on so we should be careful when introducing them.

KrooshalUX:
@ashwin-pc  I believe we have identified a path forward that satisfies your concerns. @BSFishy  can you please follow up with the approach you are going to move forward with? Do we need to close this PR?

BSFishy:
No, I'll update this one. I think the way forward is making an `icon` prop (instead of `searchIcon`), allowing it to be an icon string, boolean (true for search icon, false for no icon), or undefined (indicating no icon). Is this an agreeable solution @KrooshalUX @ashwin-pc ?

ashwin-pc:
yep that works for me :)

src/components/combo_box/combo_box.test.tsx
```diff
@@ -221,6 +221,30 @@ describe('props', () => {
...
+  test('icon is rendered', () => {
+    const component = shallow(
+      <OuiComboBox
+        options={options}
+        selectedOptions={[options[2]]}
+        icon={true}
+      />
+    );
+
+    expect(component).toMatchSnapshot();
```
ashwin-pc, line 234: The resulting snapshots int the two tests here only show that the icon prop is set, It does not test the rendered html. The goal is to make sure that the icon has rendered, so we should explicitly check for that.

ashwin-pc (commented):
The rest of the PR looks good, just update the test to explicitly check for the Icon element

ashwin-pc (approved):

src/components/combo_box/combo_box_input/combo_box_input.tsx
```diff
@@ -272,6 +275,20 @@ export class OuiComboBoxInput<T> extends Component<
...
+
+      icon = (
+        <OuiIcon
+          type={typeof iconProp === 'string' ? iconProp : 'search'}
+          size={compressed ? 's' : 'm'}
+          className={iconClasses}
+          data-test-subj="comboBoxIcon"
+        />
+      );
+    }
```
joshuarrrr, line 290: nit - you can avoid using `let` by rewriting as a function.

joshuarrrr (commented):
This implementation looks good, but I think the props of the combo box are now confusing with two icon-related props that control different icons.

AMoo-Miki (approved):

AMoo-Miki (approved):

PR #142, opened 2022-12-02
Update qs

BSFishy:
### Description
Update qs to latest patch version
 
### Check List
- [x] All tests pass
  - [x] `yarn lint`
  - [x] `yarn test-unit`

### DIFF
--- a/src-docs/src/views/badge/beta_badge.js
+++ b/src-docs/src/views/badge/beta_badge.js
@@ -20,21 +20,21 @@ export default () => (
     {colors.map((item, index) => (
       <div key={index}>
         <OuiBetaBadge
-          label="Beta"
+          label="Experimental"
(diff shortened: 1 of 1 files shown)


AMoo-Miki:
What's the motivation for this? I ask because the older `qs@~6.5.2` still remains.

BSFishy:
This is because `~6.5.2` is only allowed to upgrade for patch versions. `^6.7.3` is allowed to upgrade minor versions, so they should stay separate, as per https://docs.npmjs.com/about-semantic-versioning

ashwin-pc:
Oui dependencies arent shared like OSD's are. I dont think there is a need to follow the same semver restrictions that OSD follows here and we should be able to safely update all the packages without breaking semver

PR #165, opened 2022-12-23
Add test and lint workflow

### LINKED ISSUE #157: Add Github workflows to build and test
**Is your feature request related to a problem? Please describe.**

When a PR comes through, we have to either manually run builds/tests ourselves or trust that the PR author has run and confirmed they work correctly.

**Describe the solution you'd like**

There should be a Github workflow to run builds and tests automatically for PRs so we can confirm nothing is breaking without putting in additional time validating these manually. We can model something after [Dashboard's workflow](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/.github/workflows/build_and_test_workflow.yml).

**Describe alternatives you've considered**

Keeping the system as-is, but there as it is, time is wasted pulling the changes and running the tests manually.

BSFishy:
### Description
Adds a workflow to lint and test on pushes to branches and on PR's.
 
### Issues Resolved
#157 
 
### Check List
- [x] All tests pass
  - [x] `yarn lint`
  - [x] `yarn test-unit`

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -1,0 +1,37 @@
+name: Build and test
+
+on:
+  push:
+    branches: ['**' '!backport/**']
+    paths-ignore:
+      - '**/*.md'
+  pull_request:
+    branches: ['**']
+    paths-ignore:
+      - '**/*.md'
+
+jobs:
+  build-lint-test:
+    runs-on: ubuntu-latest
+    name: Build and Verify
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version-file: '.nvmrc'
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Setup Yarn
+        run: |
+          npm uninstall -g yarn
+          npm i -g yarn@1.22.10
+          yarn
+
+      - name: Run linter
+        run: yarn lint
+
+      - name: Run unit tests
+        run: yarn test-unit
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -26,12 +26,13 @@ jobs:
         with:
           node-version-file: '.nvmrc'
           registry-url: 'https://registry.npmjs.org'
+          cache: 'yarn'
 
       - name: Setup Yarn
         run: |
           npm uninstall -g yarn
           npm i -g yarn@1.22.10
-          yarn
+          yarn install --frozen-lockfile
 
       - name: Run linter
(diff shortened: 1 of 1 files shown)


.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,37 @@
+name: Build and test
+
+on:
+  push:
+    branches: ['**', '!backport/**']
```
joshuarrrr, line 5: also running unit tests on backports may provide some additional confidence against regressions, but I think you can also make a good case to exclude them.

joshuarrrr (dismissed):

.github/workflows/build_and_test.yml
AMoo-Miki: I am with Josh on having tests run on backports.

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,37 @@
...
+    paths-ignore:
+      - '**/*.md'
+  pull_request:
+    branches: ['**']
+    paths-ignore:
+      - '**/*.md'
+
+jobs:
+  build-lint-test:
+    runs-on: ubuntu-latest
```
AMoo-Miki, line 15: Do we know that OUI builds on Windows? Should we add a `build and Verify` for Windows too?

BSFishy: Added Windows workflow :+1:

AMoo-Miki (approved):

AMoo-Miki (approved):

.github/workflows/build_and_test.yml
```diff
@@ -0,0 +1,41 @@
...
+        with:
+          node-version-file: '.nvmrc'
+          registry-url: 'https://registry.npmjs.org'
+          cache: 'yarn'
+
+      - name: Setup Yarn
+        run: |
+          npm uninstall -g yarn
+          npm i -g yarn@1.22.10
+          yarn install --frozen-lockfile
```
ashwin-pc, line 35: This seems unnecessary, see https://github.com/actions/setup-node/blob/main/docs/advanced-usage.md#caching-packages-data. The previous action should already install and cache yarn correctly. I havent validated this, but this seems to be correct based on the documentation for `setup-node`

BSFishy: Reading through the docs and a little bit of the source code for the action, it looks like it essentially is just a shorthand to use the [`actions/cache`](https://github.com/actions/cache/blob/main/examples.md#node---npm) action on `node_modules` and `yarn.lock`, meaning that if we update dependencies or change dependencies, those changes wouldn't be reflected in the workflow.

ashwin-pc (approved):

PR #204, opened 2023-01-09
Replace @elastic/datemath with @opensearch/datemath

### LINKED ISSUE #129: Replace @elastic/datemath with OSD datemath
With our move away from dependence on Elastic, we need to remove the dependency & peer dependency on `@elastic/datemath`. Elastic themselves have moved away from using that package on NPM and instead are telling people to use the datemath library in Kibana. [OSD has its own implementation of datemath](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/b19c1717a29fffa1b40f8073c6dfb3fb4fe85972/packages/opensearch-datemath/index.js), which should be brought over to OUI. Simply copying the `index.js` and `index.d.ts` should be fine.

Action items
 - [ ] Copy the `datemath` package from OSD to OUI
 - [ ] Refactor existing usages of `@elastic/datemath` to use the OSD datemath implementation
 - [ ] Remove dependency and peer dependency on `@elastic/datemath`

> **Note:** The datemath library we copy from OSD should not be a publicly accessible interface. That means that consumers of OUI should not be able to directly access it in any way. We may change this in the future, but by not making it publicly accessible, the implementation is opaque, meaning we won't have to worry about breaking changes relating to it.

BSFishy:
Signed-off-by: Matt Provost <provomat@amazon.com>

### Description
Replaces `@elastic/datemath` with `@opensearch/datemath` from OSD. This helps with our legal compliance and will help make it easier to fix security vulnerabilities in the future.
 
### Issues Resolved
#129
 
### Check List
- [x] All tests pass
  - [x] `yarn lint`
  - [x] `yarn test-unit`


package.json
```diff
@@ -131,8 +131,8 @@
     "@babel/preset-react": "^7.10.4",
     "@babel/preset-typescript": "^7.12.1",
     "@elastic/charts": "^30.2.0",
-    "@elastic/datemath": "^5.0.3",
     "@elastic/eslint-config-kibana": "^0.15.0",
+    "@opensearch/datemath": "file:./packages/opensearch-datemath",
```
AMoo-Miki, line 135: This should be named `oui-datemath`.

ananzh: @BSFishy could you check what should be the name of this package?

AMoo-Miki: Since the names we wanted are already taken, it doesn't make sense to hold this on just a simple name which is used internally only.

AMoo-Miki (approved):

packages/opensearch-datemath/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "../../tsconfig.base.json",
+  "compilerOptions": {
+    "tsBuildInfoFile": "../../build/tsbuildinfo/packages/opensearch-datemath"
+  },
+  "include": [
+    "index.d.ts"
+  ]
+}
```
joshuarrrr, line 9: nit, missing blank line EOF

joshuarrrr (approved):
I didn't review the actual implementation of the `openserach-datemath` package, which I assume was directly forked. If there are functional changes there, let me know and I'll give it a second look.

joshuarrrr:
backported in https://github.com/opensearch-project/oui/pull/872

PR #619, opened 2023-03-22
[CCI] Optimize Mend Security Check pipeline

SergeyMyssak:
### Description
Optimized Mend Security Check pipeline
 
### Issues Resolved
Added `mend_security_check.yaml` to optimize `Mend Security Check` pipeline

### Check List
- [x] All tests pass
  - [x] `yarn lint`
  - [x] `yarn test-unit`


BSFishy:
Thanks for the contribution, this is a really cool way to not waste resources. We should make sure this follows org guidelines. I'm currently out of office, so pinging @seanneumann @ashwin-pc @joshuarrrr @AMoo-Miki

BSFishy:
> I'm wondering why we should call this pipeline when we changing `**/*.js`, `**/*.jsx`, `**/*.ts`, `**/*.tsx` files? As far as I know, it seems to only check for dependencies 🤔

My understanding is that Mend will also check source files (in some way) which is why we don't solely rely on Dependabot. Gauging from the examples (`**/*.m **/*.mm  **/*.js **/*.php **/*.rb **/*.py`), it seems like that's what Mend suggests.

SergeyMyssak:
> > I'm wondering why we should call this pipeline when we changing `**/*.js`, `**/*.jsx`, `**/*.ts`, `**/*.tsx` files? As far as I know, it seems to only check for dependencies 🤔
> 
> My understanding is that Mend will also check source files (in some way) which is why we don't solely rely on Dependabot. Gauging from the examples (`**/*.m **/*.mm **/*.js **/*.php **/*.rb **/*.py`), it seems like that's what Mend suggests.

I did some research on [Mend for GitHub.com](https://github.com/apps/mend-for-github-com). Based on its description and its official documentation, I didn't find anywhere that it provides additional functionality to analyze code for vulnerabilities. It primarily focuses on identifying vulnerabilities in open source software dependencies, and does not currently offer additional security features for analyzing the source code itself.

The reason why we specify `**/*.js, **/*.jsx, **/*.ts, **/*.tsx` files is because some projects may have additional dependencies that are not explicitly listed in the package manager files, such as scripts or libraries included in the source code files, such as `<script src="https://code.jquery.com/jquery-3.6.0.min.js"></script>`.

We can prevent the use of inline scripts or dependencies that are not explicitly listed in the package manager files by doing the following:
1) Add `no-script-url` rule to ESLint (it's currently already in use). 
- This rule checks for the use of the `javascript: pseudo-protocol` in a URL, which is often used to execute inline scripts, and reports an error if it is detected.
2) Use `eslint-plugin-import` plugin - `import/no-extraneous-dependencies` rule. 
- It checks for imports that are not explicitly listed as dependencies in the package.json file.

When I applied the changes above, my ESLint fails in several places with the following error of this type:
> 'vfile-message' should be listed in the project's dependencies. Run 'npm i -S vfile-message' to add it  import/no-extraneous-dependencies

Full list of places where there is an error "should be listed in the project's dependencies":
| Library                | Location                                                                                                                                                                                                                                                |
|------------------------|---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| history@4.10.1         | oui/src-docs/src/routes.js<br/> oui/src-docs/src/store/configure_store.js                                                                                                                                                                               |
| @babel/types@7.12.10   | oui/src-docs/src/services/playground/dummyFunction.js<br/>oui/src-docs/src/views/facet/playground.js<br/>oui/src-docs/src/views/key_pad_menu/playground.js<br/>oui/src-docs/src/views/link/playground.js<br/>oui/src-docs/src/views/panel/playground.js |
| @babel/template@7.10.4 | oui/src-docs/src/services/playground/utils.js                                                                                                                                                                                                           |
| brace@0.11.1           | oui/src-docs/src/views/code_editor/custom_mode.js<br/>oui/src-docs/src/views/code_editor/playground.js<br/>oui/src-docs/src/views/code_editor/read_only.js                                                                                              |
| cheerio@1.0.0-rc.5     | oui/src/components/icon/icon.test.tsx                                                                                                                                                                                                                   |
| vfile-message@2.0.4    | oui/src/components/markdown_editor/markdown_editor.tsx<br/>oui/src/components/markdown_editor/markdown_types.ts                                                                                                                                         |

SergeyMyssak:
Hello guys! I corrected the code of the `mend_security_check.yaml` file, now it works properly, that is, it takes `yarn.lock` file from the `main` branch and compares it to the current one.

Currently, I am getting this error:
> Error: invalid auth configuration, please run mend auth again

I think that the authentication credentials are invalid or not properly configured. Could anyone please check the values for `MEND_EMAIL` and `MEND_USER_KEY` secrets are correct and properly set in the GitHub repository settings ("Settings" -> "Secrets")?

More info - https://docs.mend.io/bundle/cli/page/login_to_the_mend_cli_with_a_script.html

BSFishy:
If all we really need to do is change the workflow to only include `yarn.lock`, could the scope of this PR change to include those ESLint changes and change this line to only include `yarn.lock`:

https://github.com/opensearch-project/oui/blob/af9324a59324857ea928fd0de3e5baf77adb8fbf/whitesource.config#L234

It doesn't really make sense to me to have 2 redundant workflows, especially if one of them requires more setup and maintenance

SergeyMyssak:
> If all we really need to do is change the workflow to only include `yarn.lock`, could the scope of this PR change to include those ESLint changes and change this line to only include `yarn.lock`:
> 
> https://github.com/opensearch-project/oui/blob/af9324a59324857ea928fd0de3e5baf77adb8fbf/whitesource.config#L234
> 
> It doesn't really make sense to me to have 2 redundant workflows, especially if one of them requires more setup and maintenance

I understand your point of view. I think it would be better to consider this PR as a proposal in which we can compare and analyze what will be better and whether we need it, then apply the discussed changes to improve the project. The aim of this proposal is to reduce the costs and resources that are allocated to Mend Security Check. It should be noted that I am not sure if my changes will help reduce costs and resources, since Mend Security Check is only available as Enterprise, I cannot test and compare it on my test project.

**Description of the two ways:**

`Mend for GitHub.com (GitHub App)` provides an easy and automated way to integrate security scans into the project's workflow without requiring any additional setup or maintenance. The integration is seamless and allows developers to receive automated pull request comments with security findings, making it easier to identify and fix vulnerabilities.

`Own pipeline in GitHub Actions` provides more customization and control over the security scanning process. It allows us to customize it, such as specifying the exact rules to be checked and providing more flexibility. We can set up different workflows for different branches and use conditional statements to run certain checks only under specific conditions. This means that we can customize our workflow to fit our project's specific needs, rather than relying on a one-size-fits-all approach. Additionally, it gives the project more visibility and control over the security scans, which can provide additional assurance to contributors and users. 

In terms of costs, I can assume that `Mend for GitHub.com` typically runs on a dedicated infrastructure, and could potentially lead to higher costs for us if we have a large codebase or run many scans. By using `GitHub Actions` there may be a need to pay for additional resources, but GitHub Actions still offers competitive pricing compared to other services. Additionally, the ability to customize and control the resources used can help optimize costs for us.

Overall, it's important to evaluate the specific needs and usage of the project to determine the most appropriate option. If the project requires a simple and automated solution, `Mend for GitHub.com (GitHub App)` may be the better option. For example, as was suggested just leave `yarn.lock` in `includes` option, but I don't know what this application does under the hood, whether it is called every time or not. If the project requires more customization and control over the security scanning process, running our own pipeline may be the better option. And as an option, we can refuse `Mend for GitHub.com (GitHub App)` and switch completely to pipelines.

I can assume the following action item - we need to understand if it will really benefit from the following points of view:
- Costs
- Resources
- Maintenance
- Advanced functionality

FYI: as an addition to our pipeline, we can also add the ability to add a comment or create a PR:
```yaml
- name: Scan code
  run: |
    mend sca --json > report.json

- name: Parse report
  id: parse-report
  run: |
    # Parse JSON report and generate formatted comment
    echo "name=comment" >> $GITHUB_ENV

- name: Create comment
  uses: actions/github-script@v4
  with:
    github-token: ${{ secrets.GITHUB_TOKEN }}
    script: |
      const { owner, repo, number } = context.issue;
      await github.issues.createComment({ owner, repo, issue_number: number, body: env.comment });
```

```yaml
- name: Create pull request
  uses: peter-evans/create-pull-request@v3
  with:
    commit-message: 'Mend SCA security findings'
    title: 'Mend SCA security findings'
    body: |
      Security vulnerabilities have been found in the codebase.
      # Some code
      Please review the findings and take appropriate actions.

    branch: 'mend-sca-findings'
    delete-branch: true
    labels: 'security'
    reviewers: 'username1,username2'
```

BSFishy:
Thanks for adding context. OUI is a much smaller project than Dashboards (we only have 2 dedicated developers on the project, including me), so I think I'll wait to see what Dashboards does to solve this problem. Of note, Mend was added to this repository by leadership as a redundant security check, so it's possible that there is an organization-wide standard which may be diverted away from by doing something like this. I'm not sure where an org-wide conversation about this might take place, but [opensearch-project/.github](https://github.com/opensearch-project/.github) might be a good place to start. In the meantime, I'll keep an eye on https://github.com/opensearch-project/OpenSearch-Dashboards/issues/3710

ashwin-pc:
@SergeyMyssak This is a really nice proposal, but I think the org's stance is not to really optimize this since its not a particularly resource intensive test and in theory should always pass. Especially since we are committing to fix all security vulnerabilities in the older versions too. So in theory, there should be no supported version of the cod where Mend should fail, and this is a way to ensure that we do that.

In reality, this is not always possible, but thats more of an exception than the norm, so I dont think its best to selectively run mend just yet.

dblock:
Looks like consensus is not to merge this PR. Should it be closed @ashwin-pc?

[Catch All Triage w/ [1](https://github.com/dblock), [2](https://github.com/krisfreedain), [3](https://github.com/andrross)]

PR #636, opened 2023-03-27
Add build artifacts to release

BSFishy:
### Description
Adds build artifacts to the release drafter workflow. By running `yarn build` before drafting the release, we can make sure releases include all of the compiled JS and CSS needed by consumers
 
### Check List
- [x] All tests pass
  - [x] `yarn lint`
  - [x] `yarn test-unit`

### DIFF
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -13,8 +13,24 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+      - name: Setup Node
+        uses: actions/setup-node@v3
+        with:
+          node-version-file: '.nvmrc'
+          registry-url: 'https://registry.npmjs.org'
+      - name: Setup Yarn
+        run: |
+          npm uninstall -g yarn
+          npm i -g yarn@1.22.10
+          yarn install --frozen-lockfile
+      - name: Generate build
+        run: |
+          yarn build
+          yarn pack --filename oui.tar.gz
       - name: Release
         uses: softprops/action-gh-release@v1
         with:
           draft: true
           generate_release_notes: true
+          files: |
+            oui.tar.gz
--- a/.github/workflows/release-drafter.yml
+++ b/.github/workflows/release-drafter.yml
@@ -27,10 +27,12 @@ jobs:
         run: |
           yarn build
           yarn pack --filename oui.tar.gz
+          mkdir oui && mv oui.tar.gz oui/oui.tar.gz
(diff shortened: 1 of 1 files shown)


joshuarrrr (dismissed):

.github/workflows/release-drafter.yml
```diff
@@ -13,8 +13,24 @@ jobs:
...
+          registry-url: 'https://registry.npmjs.org'
+      - name: Setup Yarn
+        run: |
+          npm uninstall -g yarn
+          npm i -g yarn@1.22.10
+          yarn install --frozen-lockfile
+      - name: Generate build
+        run: |
+          yarn build
+          yarn pack --filename oui.tar.gz
```
gaiksaya, line 29: Recently we release reporting0cli and they are creating the tarball using npm pack. See: https://github.com/opensearch-project/reporting-cli/blob/main/.github/workflows/release-drafter.yml#L17-L29
is that different from whet you are doing here?

.github/workflows/release-drafter.yml
```diff
@@ -13,8 +13,24 @@ jobs:
...
       - name: Release
         uses: softprops/action-gh-release@v1
         with:
           draft: true
-          generate_release_notes: true
\ No newline at end of file
+          generate_release_notes: true
+          files: |
+            oui.tar.gz
```
gaiksaya, line 36: The current pipeline unzipping whatever artifacts are inside a tarball. You might want to zip this inside one more tarball.

gaiksaya (commented):
I don't have much context on how artifacts should be packed for this project. Added some comments based on what other project did.

.github/workflows/release-drafter.yml
BSFishy: I don't think so. I think `yarn pack` passes through or emulates the functionality of `npm pack` while still being a yarn command. I think I'd prefer to stick with yarn commands as much as possible so we don't have any chance of worrying about conflicting `yarn.lock` and `package-lock.json` files

.github/workflows/release-drafter.yml
BSFishy: For clarification, you're saying that `oui.tar.gz` should be wrapped inside of another tarball?

gaiksaya: Yes

BSFishy: Done

joshuarrrr (approved):

ashwin-pc (approved):

gaiksaya (approved):

Issue #648, opened 2023-03-30
BWC Testing Proposal


BSFishy:
OpenSearch has backwards-compatibility testing which helps ensure backwards compatibility isn't broken between releases. OUI doesn't fit into that system, and thus doesn't have any sort of backwards compatibility testing as of now. I propose that we write a tool to help ensure backwards compatibility doesn't break between releases, as well as give us guidance on what branches to backport to.

To have backwards compatibility tests, we need to have a public API. Luckily, TypeScript generates an `oui.d.ts` file for us that we can use as a public API and compare between releases to check for issues. A tool could compare the two, generate some sort of diff, and give us info based on what it finds. This work _could_ be done manually, however OUI has a massive public API and automating it would save a lot of time.

The tool would follow these rough steps:
 1. Read `oui.d.ts` in and parse it into an AST
 2. Compare the [old] and [new] `oui.d.ts` files and generate a diff
 3. Write the diff to a file

Once we have a file that represents the public API diff between two versions, we can provide our 2 analyses. Our first would check if there are any changes that are breaking for a certain release type (we can have different settings for minor and patch, no major because majors are breaking). The second would analyze the diff given the same criteria and suggest a branch to backport to.

For example, when a PR comes in, the BWC tester would run and give us a list of changes. If there was no change to any public API, the bot could suggest it _could_ be backported to a patch release branch. Keyword there is _could_, as certain changes might not change the public API but still need a minor or major version bump. Likewise, if there's a backwards compatible change in the PR, it _could_ suggest backporting to a minor release branch, and a major release branch if there's backwards incompatible changes. These suggestions could be left as a comment in the PR to help reviewers.

Another workflow could run that fails if backwards compatible changes are detected. For example, this workflow wouldn't run on `main`, as backwards incompatible changes are allowed on that branch, but would run on PR's merging into `1.0` or `1.x`. This would make it more clear to reviewers if there are backwards incompatible changes in the PR.

Again, this tool would really only test the public API. It doesn't test the functionality at all. That means that functionality could change that is backwards incompatible but the tool would still flash green lights. It would be more of guiderails that help reviewers than the end-all and be-all tool to ensure backwards compatibility. However, it would still be great help for reviewers.

Please feel free to drop any questions, comments, or suggestions.

ashwin-pc:
Great proposal! This will significantly improve the confidence in backports! some of my initial thoughts on the proposal:

[...]

PR #665, opened 2023-04-05
Add build steps to workflow

BSFishy:
### Description
Adds workflow steps to run `yarn build` and `yarn build-docs`. These are good as verification steps and also generate `oui.d.ts`, which will be very useful for #648. They are scheduled to run after both `yarn test-unit` and `yarn lint` run because they are pretty hefty commands and should only be run if absolutely necessary.
 
### Check List
- [x] All tests pass
  - [x] `yarn lint`
  - [x] `yarn test-unit`


.github/workflows/build_and_test.yml
```diff
@@ -39,3 +39,33 @@ jobs:
...
+          cache: 'yarn'
+
+      - name: Setup Yarn
+        run: |
+          npm uninstall -g yarn
+          npm i -g yarn@1.22.10
+          yarn install --frozen-lockfile
+
+      - name: Build
+        run: yarn build
```
joshuarrrr, line 68: Can we skip this for PRs with no that only have docs changes, or otherwise scope it to only necessary PRs?

.github/workflows/build_and_test.yml
```diff
@@ -39,3 +39,49 @@ jobs:
...
+        with:
+          filters: |
+            src:
+              - 'src/**'
+              - package.json
+              - yarn.lock
+            docs:
+              - 'src-docs/**'
+              - package.json
+              - yarn.lock
```
BSFishy, line 79: To reviewers: should other related build files be added here? Ex. `postcss.config.js`, `tsconfig.json`, ...

joshuarrrr: yeah, I think adding those is a good idea. It's hard to know whether an allowlist or a denylist is easier in this case.

joshuarrrr (approved):

BSFishy:
Converting to draft to rethink some of the filtering stuff and also get build working on Windows

BSFishy:
I removed the path filter. I think at this point, adding a path filter would require more thought and effort than it's worth. I'd just like to get this out, and we can make a follow-up issue to figure out path filters.

.github/workflows/build_and_test.yml
AMoo-Miki: Created #828

AMoo-Miki (approved):

.github/workflows/build_and_test.yml
```diff
@@ -39,3 +39,29 @@ jobs:
 
       - name: Run unit tests
         run: yarn test-unit
+
+  build:
+    runs-on: ubuntu-latest
```
BSFishy, line 44: https://github.com/opensearch-project/oui/issues/950

joshuarrrr (approved):

Issue #693, opened 2023-04-11
[CCI] Move OuiBottomBar default inline styles to the CSS file


SergeyMyssak:
**Is your feature request related to a problem? Please describe.**

It is not possible to override the position of `OuiBottomBar` component without `!important` in the CSS file because the component uses inline styles. We have the option of not using `!important` in our styles, but then we would have to remove the default inline styles and use them in the CSS file so we have more flexibility.

**Describe the solution you'd like**

Move default inline styles to the CSS file

**Additional context**

The need for this issue was caused by https://github.com/opensearch-project/OpenSearch-Dashboards/pull/3804, and may also be caused in the future by something similar.

joshuarrrr:
@SergeyMyssak Thanks so much for opening the issue and PR over here!

SergeyMyssak:
We can close this issue as https://github.com/opensearch-project/oui/pull/694 has been merged

PR #694, opened 2023-04-11
[CCI] Move OuiBottomBar default inline styles to the CSS file

SergeyMyssak:
### Description
Moved OuiBottomBar default inline styles to the CSS file
 
### Issues Resolved
https://github.com/opensearch-project/oui/issues/693
 
### Check List
- [x] All tests pass
  - [x] `yarn lint`
  - [x] `yarn test-unit`

### DIFF
--- a/src/components/bottom_bar/__snapshots__/bottom_bar.test.tsx.snap
+++ b/src/components/bottom_bar/__snapshots__/bottom_bar.test.tsx.snap
@@ -6,7 +6,6 @@ Array [
     aria-label="aria-label"
     class="ouiBottomBar ouiBottomBar--fixed ouiBottomBar--paddingMedium testClass1 testClass2"
     data-test-subj="test subject string"
-    style="left:0;right:0;bottom:0"
   >
     <h2
       class="ouiScreenReaderOnly"
@@ -29,7 +28,6 @@ Array [
   <section
     aria-label="Page level controls"
     class="ouiBottomBar ouiBottomBar--fixed ouiBottomBar--paddingMedium"
-    style="left:0;right:0;bottom:0"
   >
     <h2
       class="ouiScreenReaderOnly"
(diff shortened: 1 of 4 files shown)


src/components/bottom_bar/bottom_bar.tsx
```diff
@@ -132,9 +132,9 @@ export const OuiBottomBar = forwardRef<
       bodyClassName,
       landmarkHeading,
       usePortal = true,
-      left = 0,
-      right = 0,
-      bottom = 0,
+      left,
+      right,
+      bottom,
```
joshuarrrr, line 137: hmm... So the properties are no longer _always_ set inline, but if these props are used the styles will be. That seems reasonable, and solves are problem in OpenSearch Dashboards. But I wonder if we should consiter getting rid of these props altogether.

SergeyMyssak: I can remove them, but I believe that these properties provide us with more opportunities for customization and control over the `OuiBottomBar`. Additionally, removing them altogether could potentially limit our options in the future

ashwin-pc: Yeah i too like the idea of removing them, but that will be a breaking change since it changes the api of the bottom bar

ashwin-pc (approved):
I dont see a reason not to make this change. Would love to hear any opposing thoughts here.

src/components/bottom_bar/__snapshots__/bottom_bar.test.tsx.snap
```diff
@@ -263,7 +252,7 @@ Array [
   <section
     aria-label="Page level controls"
     class="ouiBottomBar ouiBottomBar--fixed ouiBottomBar--paddingMedium"
-    style="left:12px;right:0;bottom:0"
+    style="left:12px"
```
joshuarrrr, line 255: Can we expand the `bottom_bar.test.tsx` test cases to also cover the other positional props (`bottom`, `right`).

joshuarrrr (commented):
Looks great! My only request is to add additional test cases or assertions to get a snapshot for when the other positional props are specified.

src/components/bottom_bar/bottom_bar.tsx
BSFishy: Lets create a follow-up issue for this and create a separate PR so that these changes can be backported

BSFishy (approved):

joshuarrrr (approved):

bandinib-amzn (approved):

rednaksi91 (approved):

Issue #707, opened 2023-04-14
[CCI] Add new functionality to OuiBottomBar to have the same width as the container element


SergeyMyssak:
**Is your feature request related to a problem? Please describe.**

We need to introduce a new functionality to the `OuiBottomBar`, which will enable it to have the same width as the container element. The rationale behind this is due to a particular problem that we have encountered - https://github.com/opensearch-project/OpenSearch-Dashboards/issues/3336 and request to consider different approaches - https://github.com/opensearch-project/OpenSearch-Dashboards/pull/3804#issuecomment-1500751853.

The issue is that the location of the `OuiBottomBar` component may not have any knowledge about its container, including its width. Since the width of the container is not fixed and can vary due to external factors, it is necessary to adjust the `OuiBottomBar` component width accordingly. Using a predefined position for the `OuiBottomBar` component in container styles is not always practical because these values can change or be dynamic.

**Describe the solution you'd like**

In order to resolve this issue, it is necessary to create a fixed-position solution for the `OuiBottomBar` that can adapt to the width of its container.

BSFishy:
One thing to note here is that React components usually don't, and moreover _shouldn't_, be able to access their containers. If a component has access to its container that breaks the "componentized" nature of React. You could resolve this by having a prop that describes how far from the edges of the screen the bottom bar should be, or in this specific case, maybe even just a boolean if the navigation bar is open. However, all in all, I think a component trying to get info about its container is a bad idea.

kgcreative:
Honestly, this may just be a case of ensuring that bottom bar is in the right container, and making sure that the width is set to 100%. I suspect we might be able to do this via CSS + layout (Which may mean revising the way we recommend we lay out certain pages), rather than trying to use absolute positioning and other tricks

KrooshalUX:
I think this is important context - https://github.com/opensearch-project/OpenSearch-Dashboards/issues/3336#issuecomment-1429010591 

Lets check the use of page template before we try to make changes to the bottom bar https://oui.opensearch.org/1.0/#/layout/page#showing-a-bottom-bar 

There is displacement to also consider, to users can reach the true end of the page as well, so abs pos makes me nervous.

ashwin-pc:
[...]

PR #708, opened 2023-04-14
[CCI] Add functionality to OuiBottomBar to have the same width as the container element

SergeyMyssak:
### Description
1) Added functionality to OuiBottomBar to have the same width as the container element.

https://user-images.githubusercontent.com/17855243/231955485-19222a05-6a9b-4fc4-874e-3870b31207f9.mov

--- 
2) Fixed the bug in `useResizeObserver` when the hook failed to immediately detect changes and consequently did not promptly return the intended result.

imgonline-com-ua-twotoone-kvwEt9fjTaP

### Issues Resolved
https://github.com/opensearch-project/oui/issues/707
 
### Check List
- [x] All tests pass
  - [x] `yarn lint`
  - [x] `yarn test-unit`

### DIFF
--- a/src/components/observer/resize_observer/resize_observer.tsx
+++ b/src/components/observer/resize_observer/resize_observer.tsx
@@ -28,7 +28,15 @@
  * under the License.
  */
 
-import { ReactNode, useCallback, useEffect, useRef, useState } from 'react';
+import {
+  MutableRefObject,
+  ReactNode,
+  RefObject,
+  useCallback,
+  useEffect,
+  useRef,
+  useState,
+} from 'react';
 import { OuiObserver } from '../observer';
 
 export interface OuiResizeObserverProps {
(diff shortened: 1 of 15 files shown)


src/components/observer/resize_observer/resize_observer.tsx
```diff
@@ -87,58 +95,90 @@ const makeResizeObserver = (
...
+  shouldObserve?: boolean;
+}
+
+export const useResizeObserver = ({
+  element,
+  elementRef,
+  elementId,
+  observableDimension = 'all',
+  shouldObserve = true,
+}: UseResizeObserverProps) => {
```
BSFishy, line 112: Changing the prototype like this is a breaking change

BSFishy (changes requested):
It looks like this implementation probably fixes the issue, however it introduces a breaking change. What that means is that Dashboards would need to introduce a breaking change (3.0) to update to OUI 2.0, which is not on the roadmap or planned for anytime soon. Ultimately, that means that Dashboards wouldn't be able to consume the fix to the issue, at which point, why are we doing it in the first place.

The fix for the issue needs to be implemented in a backwards-compatible way, such that we only need to cut a minor version bump so that Dashboards is able to pick up the changes.

All in all, I think an approach where we use a resize observer on a specific element ID has a very ugly developer experience, and somewhat breaks React's "componentized" model. I added a few more details in the original issue.

I think in its current state, this PR shouldn't be merged until these concerns are taken care of.

src/components/observer/resize_observer/resize_observer.tsx
SergeyMyssak: Yes, I know that this is a breaking change. At the moment I can see two ways around this, as only the interface of this function has changed. The first is to create a new hook, the second is to choose how it is processed depending on the container prop type. Is there another way to around this?

SergeyMyssak:
Closing this in favour of https://github.com/opensearch-project/OpenSearch-Dashboards/pull/3978

Issue #699, opened 2023-04-12
[BUG] Data Grid component renders unused empty space when number of rows per page larger than actual number of rows


curq:
**Describe the bug**
When the number of rows per page is larger than the number of input (rows), the height of the table container is adjusted to the max number of rows per page, rather than actual number of rows rendered. As a result empty unused space will be rendered. For example, when we got 100 results from the request but the number of rows per page 1000, it will render table with height that fits 1000 rows, so basically 90% of table will be empty. I made a fork of sandbox provided in [docs](https://oui.opensearch.org/1.0/#/tabular-content/data-grid) to illustrate this issue. 
This behavior was found when resolving [this issue](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/3756) in Dashboards repo.

**To Reproduce**
I have [provided a sandbox](https://codesandbox.io/s/amazing-matsumoto-rbvnwe)
1. Open sandbox or any instance of datagrid component
2. Please select the option 1000 numbers per page ( the number of actual rows of data is 100 in my example).
<img width="539" alt="scr" src="https://user-images.githubusercontent.com/67782303/231336352-069b85c5-c9de-4361-82e5-35a4735f78ab.png">
3. Scroll down to bottom and you will see that huge amount of empty space is rendered. (100 rows and space for 900 rows)

**Expected behavior**
I'm not sure is this an expected behavior or not. So I was asked by @joshuarrrr  to open this issue in my [pull request in Dashboards](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/3797#issuecomment-1503653661) repo that solves the problem that was caused by this described behavior. Here is the link to [the issue](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/3756) that shows the problem.

**Dashboards Version**
2.6.0

**Screenshots**
Here is the screenshot from sandbox, in this image we reached the end of table, but there is still a lot of empty space below ( we can notice how scrollbar barely moved). The screenshot is from the sandbox
<img width="1001" alt="image" src="https://user-images.githubusercontent.com/67782303/231334711-e9960721-cfe6-4b2b-a17d-6d61fe40bbb3.png">

PR #726, opened 2023-04-19
[Data Grid] Fix logic when pageSize larger than number of rows

curq:
### Description
There is a bug with DataGrid component, when there is less number of rows with data than number of rows per page, the DataGrid component is rendered with height equal to number of rows per page, instead of number of actual rows. This results in an empty unused space. 

For example, when there is 99 rows but the number of rows is 120, the table is rendered with height that fits 120 rows, and not 99. Screenshot:
image
The issue is even more evident when we make number of rows per page even larger (e.g. 1000 per page). I explained more in the issue https://github.com/opensearch-project/oui/issues/699 where I provided a sandbox.

After fix:
image
You might notice that the rows per page is 99 in the screenshot (actual number of rows), but the options show 120 ( they usually are provided by the developer), so I'm not sure if this okay. 
### Issues Resolved
Fixes https://github.com/opensearch-project/oui/issues/699
 
### Check List
- [x] All tests pass
  - [x] `yarn lint`
  - [x] `yarn test-unit`

### DIFF
--- a/src/components/datagrid/data_grid.tsx
+++ b/src/components/datagrid/data_grid.tsx
@@ -715,10 +715,6 @@ export const OuiDataGrid: FunctionComponent<OuiDataGridProps> = (props) => {
     ...rest
   } = props;
 
-  if (pagination) {
-    pagination.pageSize = Math.min(pagination.pageSize, rowCount);
-  }
-
   const [isFullScreen, setIsFullScreen] = useState(false);
   const [gridWidth, setGridWidth] = useState(0);
 
--- a/src/components/datagrid/data_grid.tsx
+++ b/src/components/datagrid/data_grid.tsx
@@ -715,6 +715,10 @@ export const OuiDataGrid: FunctionComponent<OuiDataGridProps> = (props) => {
     ...rest
   } = props;
 
+  if (pagination) {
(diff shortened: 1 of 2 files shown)


ashwin-pc:
@curq I dont think you should limit the options in datagrid to fix this issue, but rather treat the row limit more like `max-height` like property, where if the content exceeds the size, we show only the number of rows that we limit to, but if the rows are lesser, we adjust the height to the height of the existing rows. 

Look at the behavior of the list here and how this behaves as an example: https://playground.opensearch.org/app/management/opensearch-dashboards/objects
Screenshot 2023-04-19 at 10 24 50 AM

ashwin-pc (approved):
Nice! pulled this down and works as expected. I did notice that when you have say 51 rows and you select see 50 rows at a time, switching to the next page keeps the row height still at 50 even though there is just a single row. But Im pretty sure that we want it that way since I prefer avoiding a layout repaint given that the primary goal of the DataGrid is to improve rendering large volumes of data, and a layout repaint can be slow and expensive for such re-renders.

BSFishy (approved):

bandinib-amzn (approved):

rednaksi91 (approved):

PR #755, opened 2023-05-10
feat: vertical oriented button group

### LINKED ISSUE #659: [Feature Request] Vertical button group
**Is your feature request related to a problem? Please describe.**

Trying to add vertical button group similar to this, but couldn't find it's supported from the doc.
image

**Describe the solution you'd like**

I'd like to have `ButtonGroup` component to also support `vertical` alignment. I can see [material-ui](https://mui.com/material-ui/react-button-group/#vertical-group) have something similar.

**Describe alternatives you've considered**

We can use custom css styles but it would be nice if OUI can support this natively. 

**Additional context**

dashboards-maps project needs to add a map control button group like this:
image

The PR: https://github.com/opensearch-project/dashboards-maps/pull/372
cc @VijayanB @junqiu-lei

ruanyl:
### Description
This commit added a new prop(orientation) to the current OuiButtonGroup, the possible values are `horizontal` and `vertical`. The default value is `horizontal` which match the current UI behavior. With `vertical` prop, the button group will be oriented vertically.

Example:

image
image

### Issues Resolved
#659
 
### Check List
- [x] All tests pass
  - [x] `yarn lint`
  - [x] `yarn test-unit`


src/components/button/button_group/_button_group_button.scss
```diff
@@ -123,20 +123,43 @@
 .ouiButtonGroup--small {
   .ouiButtonGroupButton {
     border: $ouiBorderThin;
+  }
```
ashwin-pc, line 126: Why dont we want border thin for vertical?

ashwin-pc, line 126:
```suggestion
    &:not(:first-child) {
      margin-left: -1px;
    }
  }
```

ashwin-pc (commented):
@ruanyl Nice! just a few questions

src/components/button/button_group/_button_group_button.scss
ruanyl: both vertical and horizontal has `$ouiBorderThin` which set on line 125.

ruanyl: This is set on line 133 under `&:not(.ouiButtonGroup--vertical)`

KrooshalUX:
@ruanyl great question re: compressed. I'm not entirely familiar with the implementation details so apologies if this is a bit naïve -  Is there an approach where we can expand compressed beyond just a boolean? So if orientation = vertical and compressed = true then fixed width and if orientation = horizontal and compressed = true then fixed height ? 

I'm wondering if the way this was structured is why some of the other components have orientation split into different components all together.

ruanyl:
@KrooshalUX 
> So if orientation = vertical and compressed = true then fixed width and if orientation = horizontal and compressed = true then fixed height

Technically it is doable, but if orientation = vertical with fixed width could potentially problematic as the button text might overflow.

> I'm wondering if the way this was structured is  why some of the other components have orientation split into different components all together.

Do you have an example of this?

src/components/button/button_group/_button_group_button.scss
ashwin-pc: Oops yeah my bad

ashwin-pc: I should not review PR's when im sleepy. yes you are right. I did not notice that the margin properties changed are different even though the values are the same.

src/components/button/button_group/button_group.tsx
```diff
@@ -166,13 +171,24 @@ export const OuiButtonGroup: FunctionComponent<Props> = ({
     );
   }
 
+  // Compressed style can't support `vertical` orientation because compressed button group has a fixed height
```
AMoo-Miki, line 174: I don't get why this would be a blocker. A variation of `height: auto` with a fixed-width would be enough to bypass this. no?

ruanyl: @KrooshalUX what's your opinion of a vertical compressed button should look like?

joshuarrrr:
@ruanyl @KrooshalUX Is this a requirement for any 2.10 features?

ruanyl:
@joshuarrrr No, this is just an improvement

ruanyl:
@KrooshalUX This is not a blocker.

KrooshalUX:
@ruanyl oh, no , to clarify - understood its not a blocker for 2.10, but where are we blocked on finishing the issue? I am not sure if there is any outstanding decisions / need from UX ?

src/components/button/button_group/button_group.tsx
ruanyl: @KrooshalUX It would be nice to have UX suggestion on a `vertical` + `compressed` button

joshuarrrr:
I'm going to mark this as a draft for now, but @ruanyl feel free to take it out of draft status when you're able to make the updates.

src/components/button/button_group/button_group.tsx
BSFishy: Docs on `compressed`: https://oui.opensearch.org/1.2/#/forms/compressed-forms

I think the idea is to have any sort of spacing smaller than uncompressed

ruanyl:
@virajsanghvi @AMoo-Miki @ashwin-pc Finally get the PR updated, could you take another look please?

virajsanghvi:
Can you add a changelog entry?

Curious, where would this be used in OSD? Is this part of any designs?

ruanyl:
> Can you add a changelog entry?
> 
> Curious, where would this be used in OSD? Is this part of any designs?

Sure, I will update the changelog.

This change was originally introduced for maps plugin, it currently uses two buttons vertically aligned, but would be better to use a button group.

image

ruanyl:
@virajsanghvi PR updated, could you take another look please?

PR #758, opened 2023-05-15
[CCI] Allow to portale BottomBar to any referenced DOM element

SergeyMyssak:
### Description
Allow to portale BottomBar to any referenced DOM element
 
### Issues Resolved
https://github.com/opensearch-project/oui/issues/707
 
### Check List
- [x] All tests pass
  - [x] `yarn lint`
  - [x] `yarn test-unit`


scripts/jest/config.json
```diff
@@ -19,6 +19,7 @@
   },
   "setupFiles": [
     "<rootDir>/scripts/jest/setup/enzyme.js",
+    "<rootDir>/scripts/jest/setup/html_element.js",
```
SergeyMyssak, line 22: Jest uses jsdom which does not yet support the DOM function insertAdjacentElement. More information can be found at https://github.com/jsdom/jsdom/issues/1890

BSFishy: Out of curiosity, can this be done with a polyfill? Or does it need to be done like this because of Jest's infrastructure?

src/components/bottom_bar/bottom_bar.tsx
```diff
@@ -59,26 +59,27 @@ export const POSITIONS = ['static', 'fixed', 'sticky'] as const;
...
-    affordForDisplacement?: boolean;
-  },
-  {
+    insert?: OuiPortalInsert;
     /**
-     * How to position the bottom bar against its parent.
+     * Whether the component should apply padding on the document body element to afford for its own displacement height.
+     * Only works if `position` is `fixed` and `usePortal` is true.
      */
-    position: 'static' | 'sticky';
```
BSFishy, line 79: Just so I understand this correct, options are:

```ts
{
  position: 'static' | 'sticky'
}
```
or
```ts
{
  position: 'fixed',
  usePortal: ...,
  insert: ...,
  affordForDisplacement: ...,
}
```

i.e. just adding `insert` to `position: fixed`? Are there any combinations that worked previously but don't with this change?

src/components/portal/portal.test.tsx
```diff
@@ -33,13 +33,49 @@ import { mount } from 'enzyme';
...
+    const component = mount(<OuiPortal>Content</OuiPortal>);
 
     expect(component).toMatchSnapshot();
   });
+
+  it('should attach Content to body', () => {
+    mount(<OuiPortal>Content</OuiPortal>);
+
+    expect(document.body.innerHTML).toEqual('<div>Content</div>');
+  });
```
BSFishy, line 50: I see the content of these two tests are the same, can they be combined into one test or does it make more sense to keep them separated?

src/components/portal/portal.tsx
```diff
@@ -52,31 +52,45 @@ export const INSERT_POSITIONS: OuiPortalInsertPosition[] = keysOf(
...
+  { sibling: HTMLElement; position: OuiPortalInsertPosition }
+>;
 
 export interface OuiPortalProps {
   /**
    * ReactNode to render as this component's content
    */
   children: ReactNode;
-  insert?: { sibling: HTMLElement; position: OuiPortalInsertPosition };
-  portalRef?: (ref: HTMLDivElement | null) => void;
```
BSFishy, line 62: This expands what `ref` an be, not contracts, correct? I.e. with this change, `ref` can be a larger set of inputs, rather than a smaller?

src/components/portal/portal.tsx
```diff
@@ -52,31 +52,45 @@ export const INSERT_POSITIONS: OuiPortalInsertPosition[] = keysOf(
...
+    const { root, sibling, position } = insert;
+
+    // inserting within an element
+    if (root) {
+      this.portalNode = root;
+      return;
+    }
+
+    // inserting before or after an element
+    if (sibling && position) {
```
BSFishy, line 93: If we already determine that `root` is not defined, is this if statement necessary? Or is this more of a future-proofing/maintainability decision?

BSFishy (commented):
Love this implementation! A few minor comments but overall looks really good

src/components/bottom_bar/bottom_bar.tsx
SergeyMyssak: Nice catch! I wrote this a little wrong, I've rewritten this interface and I'm afraid it's going to be a breaking change

src/components/portal/portal.test.tsx
SergeyMyssak: Good idea, combined test cases

src/components/portal/portal.tsx
SergeyMyssak: Yeap, it expands

src/components/portal/portal.tsx
SergeyMyssak: Without this condition, we have some typescript errors here, and since we are manipulating the DOM directly, it would be a good idea to check for DOM elements to avoid a sudden crash

scripts/jest/config.json
SergeyMyssak: I haven't found an official polyfill for this, it's usually custom written, so this is supposed to be a polyfill, it's just in the `setup` folder, not `polyfills`

src/components/bottom_bar/bottom_bar.tsx
BSFishy: I think that given the same code before and after this change, if the functionality remains consistent, we shouldn't consider it breaking. I.e. if I don't specify `usePortal` before and after this change, it should look and function the same, and that should be fine.

SergeyMyssak: I see what you mean. The point is that we had `usePortal` and `affordForDisplacement` as defaults `true` and now they are `false`, this change leads to different behavior if these properties were not explicitly specified. This PR in itself is breaking because we allow BottomBar of any position type to be portable, which means that the behavior that was the default for fixed position should apply to other types, but in order to avoid not-expected behavior when using BottomBar, we need to remove some defaults and specify them ourselves if needed

wbeckler:
@BSFishy Are you happy with where this landed?

BSFishy:
> @BSFishy Are you happy with where this landed?

My concern with this PR is that it was meant to go into Dashboards but I believe it's breaking. @SergeyMyssak can confirm if that's the case. I spun up a Dashboards instance with this change and went to Stack Management and changed some settings to see the bottom bar, and nothing seemed to have broken, but if there's a chance that someone will break because of this, I don't want to put it into the 1.x line.

SergeyMyssak:
> > @BSFishy Are you happy with where this landed?
> 
> My concern with this PR is that it was meant to go into Dashboards but I believe it's breaking. @SergeyMyssak can confirm if that's the case. I spun up a Dashboards instance with this change and went to Stack Management and changed some settings to see the bottom bar, and nothing seemed to have broken, but if there's a chance that someone will break because of this, I don't want to put it into the 1.x line.

In the Dashboards, we have BottomBar in 3 places:
1) [advanced_settings](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/3f357b5f58109ddaf715233aad99b4e59e6b38b9/src/plugins/advanced_settings/public/management_app/components/form/form.tsx#L343)
2) [data_source_management (create_data_source_wizard)](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data_source_management/public/components/create_data_source_wizard/components/create_form/create_data_source_form.tsx#L650)
3) [data_source_management (edit_data_source_wizard)](https://github.com/opensearch-project/OpenSearch-Dashboards/blob/main/src/plugins/data_source_management/public/components/edit_data_source/components/edit_form/edit_data_source_form.tsx#L1004)

The changes only concern the default values `usePortal` and `affordForDisplacement`. If we merge this PR, we will need to open a PR in Dashboards in which we will need to explicitly specify the `usePortal` prop in items 2 and 3, because it will have no default value. No change is required for item 1. If this is a blocker for merge in `1.x`, better leave it for `2.0`.

BSFishy (approved):
New test suite passes. I think we can consider this something like "clarifying previously ambiguous functionality/requirements" based on https://github.com/opensearch-project/oui/pull/758#discussion_r1209648043

Issue #761, opened 2023-05-17
[CCI] Align deprecation mechanisms


SergeyMyssak:
**Is your feature request related to a problem? Please describe.**

In https://github.com/opensearch-project/oui/pull/759#issuecomment-1549998055, it was proposed to improve the process of depreciation of component props to make the approach more reusable and consistent with the `deprecated` high-order component.

This is a follow-up issue for the creation and consistency of depreciation mechanisms.

**Describe the solution you'd like**

Create a `useDeprecationPropWarning` hook and align it with the `deprecated` high-order component.

SergeyMyssak:
I'm closing this issue as https://github.com/opensearch-project/oui/pull/762 has been merged

PR #762, opened 2023-05-17
[CCI] Add `useDeprecatedPropWarning` and align with `deprecated` hoc

SergeyMyssak:
### Description
Added `useDeprecatedPropWarning` and aligned with `deprecated` hoc
 
### Issues Resolved
https://github.com/opensearch-project/oui/issues/761
 
### Check List
- [x] All tests pass
  - [x] `yarn lint`
  - [x] `yarn test-unit`


src/utils/deprecated/deprecated.tsx
```diff
@@ -25,3 +45,18 @@ export const deprecated = (message: string) => {
...
+}: IDeprecatedPropWarning): void => {
+  useEffect(() => {
+    if (deprecatedProp !== undefined) {
+      const formattedMessage = `The \`${deprecatedPropName}\` prop is deprecated and will be removed in v${version}`;
+      const deprecatedMessage = getDeprecatedMessage(formattedMessage);
+
+      console.warn(deprecatedMessage);
+    }
+  }, [deprecatedProp, deprecatedPropName, version]);
+};
```
BSFishy, line 62: We could use a signature that looks something like this:

```ts
useDeprecatedProps({ props: { prop1, prop2 }, version: '2.0' });
```

That way you can use `Object.keys` on the `props` prop so you don't need to pass individual `deprecatedProp` and `deprecatedPropName`. Additionally, it would make it easier to deprecate multiple props at once.

Totally up to you, I'm not going to block this PR over this :)

src/utils/deprecated/deprecated.tsx
```diff
@@ -5,14 +5,34 @@
 
 import React, { useEffect } from 'react';
 
+interface IDeprecatedComponentWarning {
+  NewComponent: React.ComponentType<any>;
```
BSFishy, line 9: Is this intentionally capitolized?

src/utils/deprecated/deprecated.tsx
```diff
@@ -5,14 +5,34 @@
...
+}: IDeprecatedComponentWarning) => {
   return <T extends React.ComponentType<any>>(Component: T): T => {
+    const deprecatedComponentName = Component.displayName || Component.name;
+    const newComponentName = NewComponent.displayName || NewComponent.name;
+
     const DeprecatedWrapper = (props: React.ComponentProps<T>) => {
       useEffect(() => {
-        console.warn(getDeprecatedMessage(message));
+        const formattedMessage = `${deprecatedComponentName} is deprecated in favor of ${newComponentName} and will be removed in v${version}.`;
+        const deprecatedMessage = getDeprecatedMessage(formattedMessage);
```
BSFishy, line 33: It may be nice to also pass in a custom deprecation message. For example, we're thinking about deprecating the current `OuiCodeEditor` in favor of `@osd/monaco`, which might require a different message than this allows

src/utils/deprecated/deprecated.tsx
```diff
@@ -5,14 +5,34 @@
...
+  deprecatedProp: any;
+  deprecatedPropName: string;
+  version: string;
+}
+
 export const getDeprecatedMessage = (message: string): string =>
   `[DEPRECATED] ${message}`;
 
-export const deprecated = (message: string) => {
+export const deprecatedComponentWarning = ({
```
BSFishy, line 22: Have we shipped a version with `deprecated` already? If so, that name should stay and be deprecated and alias to this function instead of plainly renaming it

src/components/page/page_header/page_header.test.tsx
```diff
@@ -124,4 +124,28 @@ describe('OuiPageHeader', () => {
...
+    );
+  });
+
+  it('should not console deprecation warning', () => {
+    console.warn = jest.fn();
+
+    mount(<OuiPageHeader />);
+
+    expect(console.warn).not.toHaveBeenCalled();
+  });
```
BSFishy, line 150: Maybe we could have tests for all permutations?

I.e.
 - `iconType` and `iconProps`
 - only `iconType`
 - neither

src/utils/deprecated/deprecated.tsx
SergeyMyssak: Yes, it is intentionally capitalized because it's a component. In the new commit, I decided not to pass the component, just the component name, to be more flexible

src/utils/deprecated/deprecated.tsx
SergeyMyssak: Good suggestion, now we can pass the `getMessage` function

src/components/page/page_header/page_header.test.tsx
```diff
@@ -124,4 +124,48 @@ describe('OuiPageHeader', () => {
...
+    it('should console 1 deprecation warning without repetition', () => {
+      console.warn = jest.fn();
+
+      const component = mount(<OuiPageHeader iconType="dashboardApp" />);
+      component.setProps({ iconType: 'database' });
+
+      expect(console.warn).toHaveBeenCalledTimes(1);
+      expect(console.warn).toHaveBeenCalledWith(
+        '[DEPRECATED] The `iconType` prop is deprecated and will be removed in v2.0.0'
+      );
```
BSFishy, line 138: Really like that we're testing to make sure only 1 message is logged even when props are changed

src/components/page/page_header/page_header.test.tsx
```diff
@@ -124,4 +124,48 @@ describe('OuiPageHeader', () => {
...
+      expect(console.warn).toHaveBeenCalledTimes(2);
+      results.forEach((item) =>
+        expect(console.warn).toHaveBeenCalledWith(item)
+      );
+    });
+
+    it('should not console deprecation warning', () => {
+      console.warn = jest.fn();
+
+      mount(<OuiPageHeader />);
```
BSFishy, line 166: It's possible this _could_ be deprecated sometime later down the line, but I think it's fine for now. We can worry about that later :)

src/utils/deprecated/deprecated.tsx
```diff
@@ -3,25 +3,65 @@
...
-      value: Component.displayName || Component.name,
+      value: deprecatedComponentName,
     });
 
     return DeprecatedWrapper as T;
   };
 };
+
+interface IDeprecatedPropWarning {
+  props: Record<string, any>;
```
BSFishy, line 47:
```suggestion
  props: Record<string, unknown>;
```

Haven't tested if this suggestion actually works, but it would be ideal to avoid the usage of `any`

src/utils/deprecated/deprecated.tsx
```diff
@@ -3,25 +3,65 @@
...
 
 export const getDeprecatedMessage = (message: string): string =>
   `[DEPRECATED] ${message}`;
 
-export const deprecated = (message: string) => {
+interface IDeprecatedComponentWarning {
+  newComponentName: string;
+  version?: string;
+  getMessage?: (deprecatedComponentName: string) => string;
+}
```
BSFishy, line 15: It might be more user-friendly to have the interface look something like

```ts
{
  newComponentName: string;
  version?: string;
} | {
  getMessage: () => string;
}
```

src/utils/deprecated/deprecated.tsx
```diff
@@ -3,25 +3,65 @@
...
+}: IDeprecatedComponentWarning) => {
   return <T extends React.ComponentType<any>>(Component: T): T => {
+    const deprecatedComponentName = Component.displayName || Component.name;
+
     const DeprecatedWrapper = (props: React.ComponentProps<T>) => {
       useEffect(() => {
-        console.warn(getDeprecatedMessage(message));
+        const message =
+          getMessage?.(deprecatedComponentName) ||
+          `${deprecatedComponentName} is deprecated in favor of ${newComponentName} and will be removed in v${version}.`;
```
BSFishy, line 29: `version` is nullable. Might be better to have a message if `version` has no value? Or require a value (but I think the other option would be better in case we don't have a specific version in mind)?

src/utils/deprecated/deprecated.tsx
SergeyMyssak: It works fine, I've made changes

src/utils/deprecated/deprecated.tsx
SergeyMyssak: Good idea, I've made changes

BSFishy (approved):
This looks great, thanks for these changes!

joshuarrrr (approved):

Issue #584, opened 2023-03-09
Upgrade to Webpack 5


BSFishy:
Webpack v4 is old and outdated and should be upgraded to Webpack v5. Mend already raised a PR (#545) to do this, however that PR simply upgrades the package number and doesn't actually do any migration work. It's possible that migration isn't _extremely_ necessary, however it would be good to make sure we're not using anything that's deprecated.

This issue will be somewhat a copy of the content on Webpack's migration guide with added context for some steps and with some steps removed that have already been done: https://webpack.js.org/migrate/5/

This migration should take place after #578 is merged.

When performing this migration, please make a draft PR and put each top-level step in a separate commit. This will make it easier to see what's going on while the migration is happening, catch any mistakes during the process, and diagnose issues if they arrise.

Steps:
 - [ ] Upgrade `webpack-cli` to `^5.0.1` in `package.json` and run a `yarn install`
 - [ ] Update plugins and loaders to their latest versions
    - [ ] Upgrade `babel-loader` to `^9.1.2`
    - [ ] Upgrade `style-loader` to `^3.3.1`
    - [ ] Upgrade `css-loader` to `^6.7.3`
    - [ ] Upgrade `postcss-loader` to `^7.0.2`
    - [ ] Upgrade `sass-loader` to `^13.2.0`
    - [ ] Upgrade `file-loader` to `^6.2.0`
    - [ ] Upgrade `url-loader` to `^4.1.1`
    - [ ] Upgrade `html-webpack-plugin` to `^5.5.0`
    - [ ] Upgrade `circular-dependency-plugin` to `^5.2.2`
    - [ ] Upgrade `terser-webpack-plugin` to `^5.3.7`
    - [ ] Upgrade `fork-ts-checker-webpack-plugin` to `^8.0.0`
 - [ ] Run `yarn install` then both `yarn build` and `yarn build-docs` to make sure there are no build warnings or errors
    - If there are warnings or errors in this step, consult help here
 - [ ] Set the following Webpack config options in both Webpack configs (`src/webpack.config.js` and `src-docs/webpack.config.js`):
    - [ ] `node.Buffer`: `false`
    - [ ] `node.process`: `false`
 - [ ] Run `yarn build` and `yarn build-docs` again to see if there are any errors
    - If there are warnings or errors in this step, consult help here
 - [ ] Remove the Webpack config options from the previous step (`node.Buffer` and `node.process`)
[...]

PR #765, opened 2023-05-18
Update to Webpack 5

jmetev1:
### Description
I continued from @AbhishekReddy1127 pr to upgrade webpack. It has some import problems. I fixed them. 
 
### Issues Resolved
Fixes #584 
 
### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] All tests pass
  - [ ] `yarn lint`
  - [ ] `yarn test-unit`
- [ ] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/oui/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #770, opened 2023-05-20
[CCI] Update webpack to v5

andreymyssak:
### Description
Improved version of https://github.com/opensearch-project/oui/pull/765. It would take a long time to address all the potential improvements, so I created a new PR where I applied them.

 Run the following commands again to see if there are any errors:
- [X]  yarn build
- [X]  yarn build-docs
- [X]  yarn start
- [X]  yarn test-unit
- [X]  yarn lint

### Issues Resolved
https://github.com/opensearch-project/oui/issues/584
 
### Check List
- [X] All tests pass
  - [X] `yarn lint`
  - [X] `yarn test-unit`


package.json
```diff
@@ -122,16 +120,12 @@
   "devDependencies": {
     "@faker-js/faker": "^7.6.0",
     "@axe-core/puppeteer": "^4.1.1",
-    "@babel/cli": "^7.10.5",
-    "@babel/core": "^7.11.4",
-    "@babel/plugin-proposal-class-properties": "^7.10.4",
-    "@babel/plugin-proposal-object-rest-spread": "^7.11.0",
-    "@babel/plugin-syntax-dynamic-import": "^7.8.3",
-    "@babel/plugin-transform-async-to-generator": "^7.10.4",
```
andreymyssak, line 130: These plugins are included in `@babel/preset-env`

package.json
```diff
@@ -122,16 +120,12 @@
...
-    "@babel/plugin-transform-runtime": "^7.11.0",
```
andreymyssak, line 131: It's not strictly necessary to use `@babel/plugin-transform-runtime` if we are already using `core-js` with `useBuiltIns: 'usage'` in `@babel/preset-env`

package.json
```diff
@@ -150,21 +144,19 @@
     "@types/uuid": "^8.3.0",
     "@typescript-eslint/eslint-plugin": "^4.8.1",
     "@typescript-eslint/parser": "^4.8.1",
-    "autoprefixer": "^9.8.6",
+    "autoprefixer": "^10.4.14",
     "axe-core": "^4.1.1",
-    "babel-core": "7.0.0-bridge.0",
```
andreymyssak, line 155: Removed in favor of `@babel/core`

package.json
```diff
@@ -150,21 +144,19 @@
...
     "babel-jest": "^24.1.0",
-    "babel-loader": "^8.1.0",
-    "babel-plugin-add-module-exports": "^1.0.2",
+    "babel-loader": "^9.1.2",
+    "babel-plugin-add-module-exports": "^1.0.4",
     "babel-plugin-dynamic-import-node": "^2.3.3",
-    "babel-plugin-inline-react-svg": "^1.1.1",
+    "babel-plugin-inline-react-svg": "^2.0.2",
     "babel-plugin-pegjs-inline-precompile": "^0.1.1",
-    "babel-template": "^6.26.0",
```
andreymyssak, line 163: Removed in favor of `@babel/template`

package.json
```diff
@@ -190,28 +182,28 @@
     "eslint-plugin-react": "^7.21.3",
     "eslint-plugin-react-hooks": "^4.1.2",
     "expose-gc": "^1.0.0",
-    "file-loader": "^6.1.0",
```
andreymyssak, line 193: DEPRECATED for v5 (https://v4.webpack.js.org/loaders/file-loader/)

package.json
```diff
@@ -190,28 +182,28 @@
...
-    "postcss-loader": "^4.0.1",
+    "postcss": "^8.4.23",
+    "postcss-cli": "^10.1.0",
+    "postcss-inline-svg": "^6.0.0",
+    "postcss-loader": "^7.3.0",
     "pre-commit": "^1.2.2",
     "prettier": "^2.1.2",
     "prop-types": "^15.6.0",
     "puppeteer": "^5.5.0",
-    "raw-loader": "^4.0.1",
```
andreymyssak, line 214: DEPRECATED for v5 (https://v4.webpack.js.org/loaders/file-loader/). It was replaced with `asset/source` type (https://webpack.js.org/guides/asset-modules/#replacing-inline-loader-syntax)

package.json
```diff
@@ -229,17 +221,16 @@
...
-    "style-loader": "^1.2.1",
-    "terser-webpack-plugin": "^4.1.0",
-    "typescript": "4.0.5",
-    "url-loader": "^4.1.0",
-    "webpack": "^4.44.1",
-    "webpack-cli": "^3.3.12",
-    "webpack-dev-server": "^3.11.0",
+    "style-loader": "^3.3.2",
+    "terser-webpack-plugin": "^5.3.9",
+    "typescript": "4.1.6",
```
andreymyssak, line 230: `Typescript v4.0.5` gives an error below because `@types/express-serve-static-core` uses `TypeScript v4.1` syntax, so I had to update it
> node_modules/@types/express-serve-static-core/index.d.ts:104:68 - error TS1110: Type expected.

scripts/babel/proptypes-from-ts-props/index.js
```diff
@@ -137,10 +137,10 @@ function resolveOmitToPropTypes(node, state) {
   // extract the string values of keys to remove
   const keysToRemove = new Set(
     toRemovePropTypes.arguments[0].elements
-      .map(keyToRemove =>
+      .map((keyToRemove) =>
```
andreymyssak, line 140: `prettier` added parentheses, @BSFishy can you check if they were supposed to be added? If they shouldn't, I'll revert the `prettier` changes

src/components/search_bar/search_filters.tsx
```diff
@@ -30,7 +30,7 @@
 
 import React, { Component, Fragment, ReactElement } from 'react';
 import { createFilter, SearchFilterConfig } from './filters';
-import { Query } from './query';
+import { Query } from './query/query';
```
andreymyssak, line 33: `Typescript` doesn't want to export the `Query` class from the `index.ts` for reasons I don't understand, so I had to adjust the imports

src/components/selectable/selectable_templates/selectable_template_sitewide.tsx
```diff
@@ -249,6 +249,8 @@ export const OuiSelectableTemplateSitewide: FunctionComponent<OuiSelectableTempl
       options={formattedOptions}
       renderOption={ouiSelectableTemplateSitewideRenderOptions}
       singleSelection={true}
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
```
andreymyssak, line 253: `Typescript` is great, but nowhere without pain :disappointed:, I couldn't fix it, so I had to come up with a not-so-good solution - `@ts-ignore`

> TS2322: Type '{ onFocus: (e: React.FocusEvent<HTMLInputElement>) => void; onBlur: (e: React.FocusEvent<HTMLInputElement>) => void; onInput: (e: React.FormEvent<...>) => void; ... 295 more ...; isPreFiltered?: boolean | undefined; }' is not assignable to type 'Partial<OuiSelectableSearchProps<{ className: string; prepend: {} | null | undefined; append: {} | null | undefined; icon?: OuiIconProps | undefined; avatar?: (Pick<HTMLAttributes<HTMLDivElement>, "children" | ... 251 more ... | "onTransitionEndCapture"> & CommonProps & DisambiguateSet<...> & { ...; } & { ...; }) | ...'.   Types of property 'options' are incompatible.     Type '((DisambiguateSet<OuiSelectableGroupLabelOption<{ [key: string]: any; }>, OuiSelectableLIOption<{ [key: string]: any; }>> & CommonProps & { ...; } & HTMLAttributes<...> & { ...; }) | (DisambiguateSet<...> & ... 3 more ... & { ...; }))[] | undefined' is not assignable to type '((DisambiguateSet<Pick<OuiSelectableOptionBase, "className" | "id" | "aria-label" | "data-test-subj" | "label" | "searchableLabel" | "key" | "checked" | "disabled" | "prepend" | "append" | "ref"> & HTMLAttributes<...> & { ...; } & { ...; }, CommonProps & ... 2 more ... & { ...; }> & CommonProps & { ...; } & HTMLAttr...'.       Type '((DisambiguateSet<OuiSelectableGroupLabelOption<{ [key: string]: any; }>, OuiSelectableLIOption<{ [key: string]: any; }>> & CommonProps & { ...; } & HTMLAttributes<...> & { ...; }) | (DisambiguateSet<...> & ... 3 more ... & { ...; }))[]' is not assignable to type '((DisambiguateSet<Pick<OuiSelectableOptionBase, "className" | "id" | "aria-label" | "data-test-subj" | "label" | "searchableLabel" | "key" | "checked" | "disabled" | "prepend" | "append" | "ref"> & HTMLAttributes<...> & { ...; } & { ...; }, CommonProps & ... 2 more ... & { ...; }> & CommonProps & { ...; } & HTMLAttr...'.         Type '(DisambiguateSet<OuiSelectableGroupLabelOption<{ [key: string]: any; }>, OuiSelectableLIOption<{ [key: string]: any; }>> & CommonProps & { ...; } & HTMLAttributes<...> & { ...; }) | (DisambiguateSet<...> & ... 3 more ... & { ...; })' is not assignable to type '(DisambiguateSet<Pick<OuiSelectableOptionBase, "className" | "id" | "aria-label" | "data-test-subj" | "label" | "searchableLabel" | "key" | "checked" | "disabled" | "prepend" | "append" | "ref"> & HTMLAttributes<...> & { ...; } & { ...; }, CommonProps & ... 2 more ... & { ...; }> & CommonProps & { ...; } & HTMLAttri...'.           Type 'DisambiguateSet<OuiSelectableGroupLabelOption<{ [key: string]: any; }>, OuiSelectableLIOption<{ [key: string]: any; }>> & CommonProps & { ...; } & HTMLAttributes<...> & { ...; }' is not assignable to type '(DisambiguateSet<Pick<OuiSelectableOptionBase, "className" | "id" | "aria-label" | "data-test-subj" | "label" | "searchableLabel" | "key" | "checked" | "disabled" | "prepend" | "append" | "ref"> & HTMLAttributes<...> & { ...; } & { ...; }, CommonProps & ... 2 more ... & { ...; }> & CommonProps & { ...; } & HTMLAttri...'.             Type 'DisambiguateSet<OuiSelectableGroupLabelOption<{ [key: string]: any; }>, OuiSelectableLIOption<{ [key: string]: any; }>> & CommonProps & { ...; } & HTMLAttributes<...> & { ...; }' is not assignable to type 'DisambiguateSet<Pick<OuiSelectableOptionBase, "className" | "id" | "aria-label" | "data-test-subj" | "label" | "searchableLabel" | "key" | "checked" | "disabled" | "prepend" | "append" | "ref"> & HTMLAttributes<...> & { ...; } & { ...; }, CommonProps & ... 2 more ... & { ...; }> & CommonProps & { ...; } & HTMLAttrib...'.               Type 'DisambiguateSet<OuiSelectableGroupLabelOption<{ [key: string]: any; }>, OuiSelectableLIOption<{ [key: string]: any; }>> & CommonProps & { ...; } & HTMLAttributes<...> & { ...; }' is not assignable to type '{ className: string; prepend: {} | null | undefined; append: {} | null | undefined; icon?: OuiIconProps | undefined; avatar?: (Pick<React.HTMLAttributes<HTMLDivElement>, "children" | ... 251 more ... | "onTransitionEndCapture"> & CommonProps & DisambiguateSet<...> & { ...; } & { ...; }) | (Pick<...> & ... 4 more ......'.                 Types of property 'className' are incompatible.                   Type 'string | undefined' is not assignable to type 'string'.                     Type 'undefined' is not assignable to type 'string'.

package.json
```diff
@@ -122,16 +121,12 @@
   "devDependencies": {
     "@faker-js/faker": "^7.6.0",
     "@axe-core/puppeteer": "^4.1.1",
-    "@babel/cli": "^7.10.5",
-    "@babel/core": "^7.11.4",
-    "@babel/plugin-proposal-class-properties": "^7.10.4",
-    "@babel/plugin-proposal-object-rest-spread": "^7.11.0",
-    "@babel/plugin-syntax-dynamic-import": "^7.8.3",
-    "@babel/plugin-transform-async-to-generator": "^7.10.4",
-    "@babel/plugin-transform-runtime": "^7.11.0",
```
andreymyssak, line 131: It's not strictly necessary to use `@babel/plugin-transform-runtime` if we are already using `core-js` with `useBuiltIns: 'usage'` in `@babel/preset-env`

src-docs/webpack.config.js
```diff
@@ -14,13 +14,13 @@ const HtmlWebpackPlugin = require('html-webpack-plugin');
...
+const NodePolyfillPlugin = require('node-polyfill-webpack-plugin');
 
 const getPort = require('get-port');
-const deasync = require('deasync');
 
-const { NODE_ENV, CI, WEBPACK_DEV_SERVER } = process.env;
+const { NODE_ENV, CI, WEBPACK_SERVE } = process.env;
 
-const isDevelopment = WEBPACK_DEV_SERVER === 'true' && CI == null;
+const isDevelopment = WEBPACK_SERVE === 'true' && CI == null;
```
andreymyssak, line 23: `WEBPACK_DEV_SERVER` was changed to `WEBPACK_SERVE` (https://github.com/webpack/webpack-cli/pull/2027)

src-docs/webpack.config.js
```diff
@@ -40,149 +40,133 @@ function employCache(loaders) {
...
-
-  // wait until we're done'
-  deasync.loopWhile(() => !isDone);
-
-  if (error) {
-    throw error;
-  } else {
-    return freeport;
-  }
-}
```
andreymyssak, line 186: `getPort` was not running for reasons I do not know, so I made `webpack.config.js` as a function and removed `deasync`

.eslintrc.js
```diff
@@ -108,6 +108,7 @@ module.exports = {
   plugins: ['jsx-a11y', 'prettier', 'local', 'react-hooks'],
   rules: {
     'prefer-template': 'error',
+    'import/no-unresolved': 'off',
```
andreymyssak, line 111: Eslint started not passing due to the detection of errors like the following
```
68:30 error Unable to resolve path to module './search?raw' import/no-unresolved
```
This may be due to the fact that these files were indexed, but I'm not sure. If you go into the code locally in another branch, IDE will also show the same error. I haven't found a more elegant solution to fix this error, so I turned off the rule.

src-docs/webpack.config.js
```diff
@@ -40,149 +40,133 @@ function employCache(loaders) {
...
-            }
-          : undefined,
-      }
-    : undefined,
-  node: {
-    fs: 'empty',
-  },
+          port,
+          allowedHosts: 'all',
+          historyApiFallback: true,
```
BSFishy, line 166: Why has the `devServer` setting changed so much? Did the API just change significantly? And do all the previous settings have an equivalent in the new settings?

src/components/icon/icon.test.tsx
```diff
@@ -60,7 +60,7 @@ function testIcon(props: PropsOf<OuiIcon>) {
       const onIconLoad = () => {
         component.update();
         expect(prettyHtml(component.html())).toMatchSnapshot();
-        resolve();
+        resolve(null);
```
BSFishy, line 63: What prompted this change?

src/components/search_bar/filters/field_value_selection_filter.tsx
```diff
@@ -37,7 +37,7 @@ import { OuiFilterButton, OuiFilterSelectItem } from '../../filter_group';
 import { OuiLoadingChart } from '../../loading';
 import { OuiSpacer } from '../../spacer';
 import { OuiIcon } from '../../icon';
-import { Query } from '../query';
+import { Query } from '../query/query';
```
BSFishy, line 40: Do we already export `Query` for `../query`? If not, it seems like that might be desirable. Even more so for backwards compatibility reasons?

src/components/selectable/selectable_templates/selectable_template_sitewide.tsx
BSFishy: We can create a followup issue to dive deeper into this

src/webpack.config.js
```diff
@@ -113,6 +124,6 @@ module.exports = {
   optimization: {
     minimize: isProduction,
     minimizer: [terserPlugin],
-    noEmitOnErrors: true,
+    emitOnErrors: true,
   },
```
BSFishy, line 128: In my testing, I had to add `usedExports: false` to disable some aggressive tree shaking. Did you run into the same thing at all?

.eslintrc.js
BSFishy: Is there an option to ignore query params? Maybe we could raise this to the ESLint team? Otherwise we could create a follow up issue to dive deeper into this

BSFishy (commented):
Looks pretty good, just had a few questions. Running integration tests now to verify compatibility with Dashboards

.babelrc.js
```diff
@@ -13,38 +13,37 @@ module.exports = {
...
-    }],
-    ["@babel/typescript", { isTSX: true, allExtensions: true }],
-    "@babel/react"
+  presets: [
+    [
+      '@babel/preset-env',
+      {
+        // `targets` property set via `.browserslistrc`
+        useBuiltIns: process.env.NO_COREJS_POLYFILL ? false : 'usage',
+        corejs: 3,
```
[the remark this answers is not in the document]
BSFishy: Just confirmed, a warning pops up when this line isn't changed:

```
WARNING (@babel/preset-env): The `corejs` option only has an effect when the `useBuiltIns` option is not `false`
```

scripts/babel/proptypes-from-ts-props/index.js
andreymyssak: I just enabled `Prettier` in the settings of my `IDE`, it seems that `Prettier` was not applied to some files in the script folder before, f.e.
```
scripts/babel/fetch-i18n-strings.js
scripts/babel/react-docgen-typescript.js
scripts/eslint-plugin/forward_ref_display_name.js
scripts/eslint-plugin/i18n.js
```
and etc.

Screenshot 2023-05-29 at 09 36 46

.babelrc.js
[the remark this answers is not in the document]
andreymyssak: > corejs: process.env.NO_COREJS_POLYFILL ? undefined : 3,

I'm not sure about changing `false` to `undefined` because there is no difference between them. According to the documentation, the default value for `useBuiltIns` is `false` (https://babeljs.io/docs/babel-preset-env#usebuiltins). I guess it is worth leaving `false` because using values that are expected by the field will help avoid surprise issues in the future.

> I think because it doesn't want the corejs version unless 'usage' is specified for useBuiltIns. Did you run into the same thing at all?

Yes, I get the same warning message. The reason for this is that we purposely do not include polyfill imports in our build (see https://github.com/opensearch-project/oui/blob/main/scripts/compile-oui.js#L190-L219). As far as I understand, this solution was inherited from here https://github.com/elastic/eui/pull/1982.

src-docs/src/views/package/changelog.js
```diff
@@ -14,7 +14,7 @@ import React from 'react';
 import { OuiMarkdownFormat } from '../../../../src';
 import { GuidePage } from '../../components/guide_page';
 
-const changelogSource = require('!!raw-loader!../../../../CHANGELOG.md').default.replace(
+const changelogSource = require('../../../../CHANGELOG.md?raw').replace(
   /## \[`master`\].+?##/s, // remove the `master` heading & contents
```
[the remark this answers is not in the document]
andreymyssak: https://github.com/opensearch-project/oui/issues/797

src-docs/webpack.config.js
andreymyssak: 1. the `contentBase`, ..., `watchOptions` were removed in favor of the static option (https://github.com/webpack/webpack-dev-server/blob/master/CHANGELOG.md#400-beta0-2020-11-27, https://github.com/webpack/webpack-dev-server/blob/master/CHANGELOG.md#static)
2. the `disableHostCheck` and `allowedHosts` options were removed in favor of the `firewall` option (https://github.com/webpack/webpack-dev-server/blob/master/CHANGELOG.md#400-beta0-2020-11-27)
3. rename `firewall` option to `allowedHosts` option (https://github.com/webpack/webpack-dev-server/blob/master/CHANGELOG.md#400-rc0-2021-07-19)

The other fields have not changed

src/components/icon/icon.test.tsx
andreymyssak: Updating `Typescript` to v4.1.6 prompted this change (see https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-1.html#resolves-parameters-are-no-longer-optional-in-promises), but I can use `Promise<void>` instead of `resolve(null)`, thanks for catching this

src/components/search_bar/filters/field_value_selection_filter.tsx
andreymyssak: We have `index.ts`, but it doesn't work for class imports and that's weird

https://github.com/opensearch-project/oui/blob/ce8acf41cc58cccbef180d1d93e0025e29e4a312/src/components/search_bar/query/index.ts#L31-L32

```bash
WARNING in ../../src/components/search_bar/search_bar.tsx 256:39-44
export 'Query' (imported as 'Query') was not found in './query' (possible exports: AST)
 @ ../../src/components/search_bar/index.ts 31:0-49 31:0-49 31:0-49
 @ ../../src/components/index.js 94:0-49 94:0-49 94:0-49
 @ ./routes.js 56:0-56 205:44-60
 @ ./index.js 42:0-30 58:28-47
```

src/webpack.config.js
andreymyssak: In the `webpack v5` migration, nothing was written about `usedExports`. If I'm not mistaken, it was also included by default for productions in version 4 (https://v4.webpack.js.org/configuration/optimization/#optimizationusedexports), so I'm not sure if we should turn it off.

.eslintrc.js
andreymyssak: As it turned out, `eslint-import-resolver-webpack` only supports "synchronous" Webpack configs (https://github.com/import-js/eslint-plugin-import/tree/main/resolvers/webpack). How critical is it to use `getPort`? Is it possible to just put port `8030`?

andreymyssak:
> Also [2086980](https://github.com/opensearch-project/oui/commit/20869800a42d1f658b3a2d4bde4d9a7a4df13296) and [e9a63af](https://github.com/opensearch-project/oui/commit/e9a63afd05e4bb532f30fd9359a71704e15a8ded) need to be reverted, as we want to backport this to `1.x` :)

If I revet these commits, then my changes won't go into `main`. Should I create another PR where I just do the cherry-pick and the necessary changes for 1.x?

andreymyssak:
## Summary of Work Done

### Dependencies
- Bumped `webpack` from v4 to v5
- Bumped `postcss` from v7 to v8
- Bumped `babel` minor version
- Bumped `typescript` from v4.0.5 to v4.1.6 (https://github.com/opensearch-project/oui/pull/770#discussion_r1199626632)
- Removed  some of the `resolutions` that were covered by a dependency update

### Formatting
`Prettier` was not applied to some files in the script folder before, so the changes can be found in the following files (https://github.com/opensearch-project/oui/pull/770#discussion_r1199626994)
```
scripts/babel/fetch-i18n-strings.js
scripts/babel/react-docgen-typescript.js
scripts/eslint-plugin/forward_ref_display_name.js
scripts/eslint-plugin/i18n.js
```

### Questions Requiring Further Discussion
1.  Can we directly use port `8030` instead of utilizing `getPortSync` and `getPort.makeRange`?

During our migration to `webpack v5`, we ran into a problem with `deasync`, as it stopped working. Previously, it was used to convert our `webpack.config.js` from asynchronous to synchronous. However, since `webpack` supports asynchronous configurations, I've converted our config file accordingly. However, this change introduced a bottleneck.

As part of our adherence to best practices, the legacy `!!raw-loader!` syntax has been replaced with `?raw`, such as:
```
const basicSource = require('!!raw-loader!./basic_window_event');
// is now
const basicSource = require('./basic_window_event?raw');
```
You can read more about this transition in the [Webpack documentation](https://webpack.js.org/guides/asset-modules/#replacing-inline-loader-syntax).

However, we ran into a problem: `eslint` does not recognize `resourceQuery`. This is because `eslint-import-resolver-webpack` only supports synchronous `webpack` configurations ([see here](https://github.com/import-js/eslint-plugin-import/tree/main/resolvers/webpack)), while we are using an asynchronous one. Since `deasync` is no longer working in webpack.config.js, we can't go back to it.

To address this, we could potentially switch back to a synchronous configuration in our `webpack.config.js`. This could be accomplished by eliminating the use of `getPortSync` and using port `8030` directly. Would it be possible to simply use port `8030` instead of `getPortSync` and `getPort.makeRange`?

2. TypeScript refuses to export the `Query class` from `index.ts` for reasons that remain unclear. See the specific issue in the [[code here](https://github.com/opensearch-project/oui/pull/770/files#diff-ea36bc8a4b8cf440fa660f1611b3c8433a68792648c24b7c4cca1dff46bc40f6R33)].

The system throws the following warning message:
```bash
WARNING in ../../src/components/search_bar/search_bar.tsx 256:39-44
export 'Query' (imported as 'Query') was not found in './query' (possible exports: AST)
 @ ../../src/components/search_bar/index.ts 31:0-49 31:0-49 31:0-49
 @ ../../src/components/index.js 94:0-49 94:0-49 94:0-49
 @ ./routes.js 56:0-56 205:44-60
 @ ./index.js 42:0-30 58:28-47
```
There is a chance that upgrading `TypeScript` might fix this problem.

3. Running into a complex type conflict with `searchProps` in `OuiSelectable`

We've encountered a type conflict problem with `searchProps` in the `OuiSelectable` component that is proving difficult to resolve. Details about the specific type error can be found [here](https://github.com/opensearch-project/oui/pull/770#discussion_r1199627886). There is a chance that upgrading `TypeScript` might fix this problem.

4. Build process displays `WARNING (@babel/preset-env): The corejs option only has an effect when the useBuiltIns option is not false`. For more details, please refer to the discussion in this [thread](https://github.com/opensearch-project/oui/pull/770#discussion_r1200768876).

5. Including `usedExports` in `webpack.config.js`. For more details, please refer to the discussion in this [thread](https://github.com/opensearch-project/oui/pull/770/files#r1200794532)

AMoo-Miki:
Thanks guys; this is a big deal and what you have done is awesome. I will pull this down and play with it a bit.

scripts/babel/proptypes-from-ts-props/index.js
AMoo-Miki: i am glad to see you are using a real IDE... unlike some people... 😉 at Ashwin.

.babelrc.js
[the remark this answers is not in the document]
AMoo-Miki: Can any of you help me with why we need any polyfills?

BSFishy:
> Thanks guys; this is a big deal and what you have done is awesome. I will pull this down and play with it a bit.

Additionally, make sure to test with Dashboards. I was running into issues where one of the auxiliary bundles wasn't exporting anything, causing issues in Dashboards.

joshuarrrr:
@andreymyssak I took a shot at resolving the conflicts, but take a look at https://github.com/opensearch-project/oui/pull/770/commits/22fb706223e0a6d73a363b931dcb224ad3412b15 to see if they make sense.

.eslintrc.js
joshuarrrr: > As it turned out, `eslint-import-resolver-webpack` only supports "synchronous" Webpack configs (https://github.com/import-js/eslint-plugin-import/tree/main/resolvers/webpack). How critical is it to use `getPort`? Is it possible to just put port `8030`?

I'm pretty sure the reason for `getPort` is that you can actually have multiple instances of OUI running on different ports - you can replicate by doing `yarn start` in multiple terminal tabs/windows, and the port will auto increment without collisions.

BSFishy:
Thinking about this more, could it be possible that the `sass-vars-to-js-loader` loader is using some sort of Webpack 4 structure and the reason why the charts module doesn't export anything?

.babelrc.js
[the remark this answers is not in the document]
AMoo-Miki: I checked the code of `babel/preset-env` and @BSFishy's solution is a valid one.

AMoo-Miki:
A few observations, after rebasing and resolving conflicts:
1. The updated `autoprefixer` causes duplicate CSS. I am not worried about this because i will be porting some changes from OSD that will eliminate the need for `autoprefixer`.
2. `*.min.css` files are no longer minified. `postcss` and `cssnano` have both been bumped but I didn't dig to see which portion is breaking the logic.
3. `oui.js` is 1.5MB larger; doesn't sound right and more investigation is needed.
4. `oui.min.js`, `oui.min.js.LICENSE.txt`, `oui.min.js.map` fail to generate.
5. `*ui_charts_theme.js` are generated as folders and not files.

AMoo-Miki:
Fixed the minification:
* Had to disable `cssDeclarationSorter` because it could have been unsafe.
* The new `cssnano` merges back-to-back media queries and removes the media queries for the lower boundary; verified that it was safe.
* The new `cssnano` also doesn't break and combine neighboring definitions.
* `rgba` is converted to `hsla`; shouldn't be a big deal.
* `svgo` produces slightly longer assets.
*

BSFishy:
> * `svgo` produces slightly longer assets.

But the assets generated by `svgo` should all be in separate files right? All OUI icons are loaded in async, and won't be in the main bundle, right?

AMoo-Miki:
> > * `svgo` produces slightly longer assets.
> 
> But the assets generated by `svgo` should all be in separate files right? All OUI icons are loaded in async, and won't be in the main bundle, right?

OUI inlines them :/

AMoo-Miki:
In an attempt to dig deeper into why the JS files were larger with webpack5, i split the `eui_charts_theme.js` files, before and after. I see a lot more imports from `corjs` included with webpack5; 57 files totaling 100KB. These include stuff like:
```
../node_modules/core-js/internals/array-from.js
../node_modules/core-js/internals/iterator-define.js
../node_modules/core-js/internals/regexp-exec.js
../node_modules/core-js/modules/es.symbol.constructor.js
```

I suspect the same is happening with `oui.js` too.

AMoo-Miki:
With `oui.js`:
* 1.5MB was added.
* 197 imports of 461KB are removed.
   * some removed imports are `index.ts` and `types.ts` files that only re-exported stuff:
      ```
      ./components/delay_hide/index.ts
      ./services/color_picker/index.ts
      ./services/popover/index.ts
      ```
* 62 imports of 214KB are added.
   * imports from `assert` make up 114KB of that.
   * 15 are from `corejs` worth 20KB.
* 1919 files have some kind of change in them, resulting in the 1.7MB delta:
   * Numerous files (i stopped counting at 50 of the first 100) have gained 5KB-10KB, all because of lines like:
   ```
   /* harmony import */ var core_js_modules_es_array_slice_js__WEBPACK_IMPORTED_MODULE_21__ = __webpack_require__(/*! core-js/modules/es.array.slice.js */ "../node_modules/core-js/modules/es.array.slice.js");
   /* harmony import */ var core_js_modules_es_array_slice_js__WEBPACK_IMPORTED_MODULE_21___default = /*#__PURE__*/__webpack_require__.n(core_js_modules_es_array_slice_js__WEBPACK_IMPORTED_MODULE_21__);
   ```

AMoo-Miki:
by the way, i couldn't get the docs to work. `yarn start` shows a blank page.

BSFishy:
> by the way, i couldn't get the docs to work. `yarn start` shows a blank page.

From CI, it looks like there's still reference to `raw-loader`: https://github.com/opensearch-project/oui/actions/runs/6177159690/job/16767685139?pr=770#step:5:31

PR #775, opened 2023-05-23
Fix combo box onFocus bug

### LINKED ISSUE #696: [BUG] Refreshing option list onFocus or onClick of ComboBox
_This issue may be lacking technical implementation detail as I am filing on behalf of a contractor using passed-on messages via Slack. This is not a live project, but a project in development and implementation issues are popping up when trying to use ComboBox within the outlined user scenario._

User scenario: During the set up of an alert monitor using the quick create flow, I see an option to send alerts to a channel. However, I do not yet have the correct channel configured, so I click "Manage Channels" and I see an interface where I can configure a new channel. When I return to the alerting quick-create flow, I expect that my newly configured channel will appear in the ComboBox menu. On click, I (should) see the ComboBox UI change to a loading state while the list of configured channels is loaded, and see the list that includes my newly configured channel. 

Screenshot 2023-04-11 at 1 38 06 PM

------
Notes from the engineer trying to implement this:
 - Trying to use onClick or onFocus to refresh the channel list is causing an issue where the user would need to click the ComboBox twice. This is not the desired experience , the user should only have to click once. 
 - This appears to be an issue that was present in the version of EUI that OUI is forked from (EUI 34.6)

BSFishy:
### Description
Previously, if `isLoading` was set in `onFocus`, options wouldn't be added when they were clicked. The issue stemmed from `onFocus` not being removed from the list of props passed to the root element of combo box. Essentially, `onFocus` would be called twice, once from the root element and once from `OuiComboBoxInput`. This meant that if you clicked an option, the root element's `onFocus` would fire, cancelling the option selection.

This bug is super nuanced so if you need more context, I can add it. Basically, just taking it out of the `rest` object fixes the bug.

### Issues Resolved
Fixes #696

### Check List
- [x] All tests pass
  - [x] `yarn lint`
  - [x] `yarn test-unit`

### DIFF
--- a/src/components/combo_box/combo_box.tsx
+++ b/src/components/combo_box/combo_box.tsx
@@ -558,6 +558,10 @@ export class OuiComboBox<T> extends Component<
   };
 
   onComboBoxFocus: FocusEventHandler<HTMLInputElement> = (event) => {
+    if (this.state.hasFocus) {
+      return;
+    }
+
     if (this.props.onFocus) {
       this.props.onFocus(event);
     }
@@ -951,6 +955,7 @@ export class OuiComboBox<T> extends Component<
       onBlur,
       onChange,
       onCreateOption,
+      onFocus,
       onSearchChange,
       options,
       placeholder,


src/components/combo_box/combo_box.tsx
```diff
@@ -558,6 +558,10 @@ export class OuiComboBox<T> extends Component<
   };
 
   onComboBoxFocus: FocusEventHandler<HTMLInputElement> = (event) => {
+    if (this.state.hasFocus) {
+      return;
+    }
```
BSFishy, line 563: This change makes it so that `onFocus` isn't called if combo box is already in focus. Basically, `onFocus` won't be called if you click on combo box again if it's already open.

cc @KrooshalUX for UX guidance on this change

ashwin-pc (approved):

src/components/combo_box/combo_box.tsx
BSFishy: Demo'd to @KrooshalUX and got approval for this change.

joshuarrrr (approved):

PR #778, opened 2023-05-23
Fix breadcrumb styling with a single breadcrumb

### LINKED ISSUE #771: [BUG] New breadcrumb design breaks with single breadcrumb
**Describe the bug**

With the new breadcrumb design (`main` and `1.x` branches), the design breaks when there is only one breadcrumb. This should _probably_ just be a case of fixing the styling.

__This needs to get out for `1.2`__

**To Reproduce**
1. Create a breadcrumb with only one crumb

**Expected behavior**
The design to not be broken

**Screenshots**

Screenshot (36)

BSFishy:
### Description
Fixes breadcrumb styling when there is only a single breadcrumb.

Before:

Screenshot (36)

After:

Screenshot (42)

### Issues Resolved
Fixes #771 
 
### Check List
- [x] All tests pass
  - [x] `yarn lint`
  - [x] `yarn test-unit`

### DIFF
--- a/src/components/breadcrumbs/_breadcrumbs.scss
+++ b/src/components/breadcrumbs/_breadcrumbs.scss
@@ -98,7 +98,8 @@
     margin-bottom: $ouiSizeXS; /* 1 */
   }
 
-  &:not(.ouiBreadcrumbWrapper--last) {
+  &:not(.ouiBreadcrumbWrapper--last),
+  &.ouiBreadcrumbWrapper--first.ouiBreadcrumbWrapper--last {
     margin-right: $ouiBreadcrumbSpacing;
   }
 }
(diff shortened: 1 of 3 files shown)


ashwin-pc:
@BSFishy is there a way we can add screenshots easily for such PR's?

ashwin-pc (approved):

BSFishy:
There's not an _easy_ way with our current tooling. I can get a screenshot

joshuarrrr:
Yeah +1 to updating our PR template to require screenshots for UI changes.

src/components/breadcrumbs/_breadcrumbs.scss
```diff
@@ -110,6 +111,10 @@
   margin-bottom: $ouiSizeXS; /* 1 */
 }
 
+.ouiBreadcrumbWall--single {
```
joshuarrrr, line 114: why the additional class and not just

```suggestion
.ouiBreadcrumbWrapper--first.ouiBreadcrumbWrapper--last {
```

BSFishy: It gets applied to different elements. You can see where in `src/components/header/__snapshots__/header.test.tsx.snap`. If the background is applied to wrapper, it will get skewed as well, which we don't want for this background.

joshuarrrr: Ah, thanks, yeah the snapshot is helpful in this case. I'm still convinced there are additional simplifications we can make here in the future, but this makes sense for now.

joshuarrrr (approved):

PR #838, opened 2023-06-20
Add workflow to cleanup caches

### LINKED ISSUE #211: Auto-delete old caches
Utilize [the workflow in this guide](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#force-deleting-cache-entries) to create a workflow that automatically deletes caches for pull requests that have been merged or closed. Once a pull request has been merged or closed, it has no use for the cache anymore, so the cache can be deleted.

BSFishy:
### Description
Adds a workflow that cleans up old caches when a PR is merged. Follows the example in these docs: https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#force-deleting-cache-entries

### Issues Resolved
Closes #211

### Check List
- [x] All tests pass
  - [x] `yarn lint`
  - [x] `yarn test-unit`

### DIFF
--- a/.github/workflows/cleanup-cache.yml
+++ b/.github/workflows/cleanup-cache.yml
@@ -1,0 +1,33 @@
+name: Cleanup caches by a branch
+on:
+  pull_request:
+    types:
+      - closed
+
+jobs:
+  cleanup:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v3
+
+      - name: Cleanup
+        run: |
+          gh extension install actions/gh-actions-cache
+
+          REPO=${{ github.repository }}
+          BRANCH="refs/pull/${{ github.event.pull_request.number }}/merge"
+
+          echo "Fetching list of cache key"
+          cacheKeysForPR=$(gh actions-cache list -R $REPO -B $BRANCH | cut -f 1 )
+
+          ## Setting this to not fail the workflow while deleting cache keys.
+          set +e
+          echo "Deleting caches..."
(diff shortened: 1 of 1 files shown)


AMoo-Miki:
If we are not using the cache, let's just stop caching by removing it [from here](https://github.com/opensearch-project/oui/blob/main/.github/workflows/build_and_test.yml#L29).

BSFishy:
> If we are not using the cache, let's just stop caching by removing it [from here](https://github.com/opensearch-project/oui/blob/main/.github/workflows/build_and_test.yml#L29).

We're using the cache. Basically it looks like this:

```
PR raised
 |
 V
Build and test workflow started
 |
 V
Cache doesn't exist, run a new `yarn install`
 |
 V
Build and test with packages
 |
 V
Cache packages
```

Then, when a change is made in the PR:
```
Change is made
 |
 V
Build and test workflow started
 |
 V
Cache is found, use cached packages (theoretically faster than installing from scratch)
 |
 V
Build and test with packages
```

At no point do we actually remove these caches, so they're left as zombies after the PR is closed. This PR adds a workflow that will delete those zombie caches, saving space for caches for new PRs or for long-standing branches like `main` or `1.x`.

AMoo-Miki:
The caches expire after 7 days of inactivity. Us deleting the cache after a PR is merged would just make it unavailable to the next run. So the cache that was created in one PR which could have benefited another PR is removed after the first one is merged; the second one does not benefit.

I think we should just let GH handle the TTL of the caches.

BSFishy:
We'll leave it up to Github then :)

PR #601, opened 2023-03-16
Docs: Add color palettes in use across OpenSearch Dashboards 

joshuarrrr:
### Description
and plugin visualizations.

Seeing these color palettes alongside the OUI color palettes and usage guidelines will help us consolidate and make visualizations more cohesive.

Includes link to various hard-coded color values and programmatic color generators.
 
### Issues Resolved

fixes https://github.com/opensearch-project/OpenSearch-Dashboards/issues/2586
 
### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [ ] All tests pass
  - [ ] `yarn lint`
  - [ ] `yarn test-unit`
- [ ] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/oui/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #856, opened 2023-06-27
Add new "middle-out" order prop option to `OuiPaletteColorBlind'

joshuarrrr:
### Description

When dealing with a large number of rotations, you may not always want to start with the darkest variations. This order makes sure the darkest and lightest variants come last.
Screen Shot 2023-06-27 at 4 29 18 PM
Screen Shot 2023-06-27 at 4 29 03 PM

### Check List
- [ ] All tests pass
  - [ ] `yarn lint`
  - [ ] `yarn test-unit`


BSFishy:
Would like to get @KrooshalUX opinion on this

joshuarrrr:
To provide a little more context - there are some legacy vis color utils in OpenSearch Dashboards that try to generate an arbitrary/infinite number of categorical colors. In practice, this generation is not done very well, and very quickly it starts generating very similar colors that are not differentiable:
Screen Shot 2023-06-28 at 5 34 51 PM (from https://github.com/opensearch-project/oui/pull/601)

I believe that the long term solution is to focus on differentiability for categorical colors, and to simply repeat/re-use categorical colors rather than using technically distinct ones.

But for now, the easiest way to at least use OUI colors as a source of truth was to leverage the `rotation` ability of `OuiPaletteColorBlind` to generate arbitrarily large palettes. The problem, however, is that by default, using the `append` order, the first 10 colors assigned would be those from the darkest rotation and the last would be those from the lightest rotation, which we know to be the "worst" colors in the total palette. Because order of assignment matters, we want a way to start with the best colors (non-rotated), and use the worst colors last.

BSFishy:
The more I'm seeing super dark or super light colors on Dashboards, the more I'm thinking I'd like this

joshuarrrr:
I'll leave this as a possibility, but leaning more toward https://github.com/opensearch-project/OpenSearch-Dashboards/issues/4697

joshuarrrr:
I'm taking this out of draft, because I'm still convinced it's useful for visualizations with >10 series. Particularly when stacking bar or area charts, it's common to sort the series so that the series with the largest magnitude come first. In this common scenario, you want the primary vis colors to come first and then start using variants.

AMoo-Miki (approved):

ashwin-pc (approved):

PR #871, opened 2023-07-01
Fix `autofill` text color in dark themes

### LINKED ISSUE #870: [BUG] In dark mode, auto-populated fields are tough to see
**Describe the bug**

Input fields that are auto-populated by the browser, like a saved usernames, are too dark to see in the dark mode for both default and Next themes. That is because the "shades" are reversed in dark mode where `$ouiColorLightestShade` is in fact nearly black and `$ouiColorDarkestShade` is nearly while. The code below wrongly uses `$ouiColorLightShade`, which is a dark color, on a dark background. 

https://github.com/opensearch-project/oui/blob/0365b429fd7935ef8b8e5b31f232c4e6a73c67e1/src/themes/oui-next/global_styling/mixins/_form.scss#L265-L277

`$ouiColorDarkestShade` or `$ouiColorDarkShade` are much more suitable for the dark themes input-fields.

**Screenshots**

Screenshot 2023-06-30 at 8 40 41 PM

AMoo-Miki:
### Description

Input fields that are auto-populated by the browser, like a saved usernames, are too dark to see in the dark mode for both default and Next themes. That is because the "shades" are reversed in dark mode where `$ouiColorLightestShade` is in fact nearly black and `$ouiColorDarkestShade` is nearly while. This change makes use of `$ouiColorDarkestShade` in both light and dark themes.

Before change:

After change:

### Issues Resolved

Fixes #870

### Check List
- [X] All tests pass
  - [ ] `yarn lint`
  - [ ] `yarn test-unit`
- [X] Update [CHANGELOG.md](./../CHANGELOG.md)

### DIFF
--- a/src/global_styling/mixins/_form.scss
+++ b/src/global_styling/mixins/_form.scss
@@ -262,16 +262,21 @@
 
     // Needs to be set for autofill
     // sass-lint:disable-block no-vendor-prefixes
-    &:-webkit-autofill {
-      -webkit-text-fill-color: lightOrDarkTheme($ouiColorDarkestShade, $ouiColorLightShade);
+    &:-webkit-autofill,
+    &:autofill {
+      /* Many browsers use `!important` in their built-in `:-webkit-autofill` style declarations,
+       * making their colors non-overridable. `-webkit-text-fill-color` is able to overwrite the
+       * appearance of texts and is used here as a workaround.
+       */
+      -webkit-text-fill-color: $ouiColorDarkestShade;
 
       ~ .ouiFormControlLayoutIcons {
-        color: lightOrDarkTheme($ouiColorDarkestShade, $ouiColorLightShade);
+        color: $ouiColorDarkestShade;
       }
 
       /* OUI -> EUI Aliases */
       ~ .euiFormControlLayoutIcons {
-        color: lightOrDarkTheme($ouiColorDarkestShade, $ouiColorLightShade);
+        color: $ouiColorDarkestShade;
       }
       /* End of Aliases */
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### 📈 Features/Enhancements
 - Rename the aliased theme files ([#863](https://github.com/opensearch-project/oui/pull/863))
+- Fix `autofill` text color in dark themes ([#871](https://github.com/opensearch-project/oui/pull/871))
 
 
 ### 🐛 Bug Fixes
--- a/src/themes/oui-next/global_styling/mixins/_form.scss
+++ b/src/themes/oui-next/global_styling/mixins/_form.scss
@@ -20,7 +20,6 @@
   $iconSize: $ouiSize;
   $iconPadding: $ouiFormControlPadding;
   $marginBetweenIcons: $ouiFormControlPadding / 2;
-
   @if ($compressed) {
     $iconPadding: $ouiFormControlCompressedPadding;
   }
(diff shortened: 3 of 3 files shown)


KrooshalUX:
@AMoo-Miki  @joshuarrrr  @BSFishy  my understanding is that any text should use either `ouiColorGhost` or `ouiColorInk`, and not any of the shades. Is that correct? If so, can we align to that vs using `ouiColorDarkestShade` as it is a different value than `ouiColorGhost` - which is what I believe the rest of the application text is using (?).

joshuarrrr:
> my understanding is that any text should use either ouiColorGhost or ouiColorInk, and not any of the shades. Is that correct?

No, I don't believe that's correct. From the source code, here's how text colors are defined.

The default variables for text are initially defined as:

https://github.com/opensearch-project/oui/blob/b80c779ba810c11078d745b3ca880c35957a5210/src/global_styling/variables/_colors.scss#L42-L46

Those exact same definitions are duplicated in our base OUI Next theme:

https://github.com/opensearch-project/oui/blob/54d50114fbf78b46d442e399055f1d8b47b5265a/src/themes/oui-next/global_styling/variables/_colors.scss#L42-L46

In dark theme variations (both the current dark theme and in OUI next), we explicitly override `$ouiTextColor` and `$ouiTitleColor` definitions to use `#DFE5EF`. Note that this is distinct from `$ouiColorGhost` and `$ouiColorInk` (`#FCFEFF` and `#0A121A`, respectively, in OUI Next Dark), as well as `$ouiColorDarkestShade` (`#DFE3E8` in OUI Dark Next):

https://github.com/opensearch-project/oui/blob/54d50114fbf78b46d442e399055f1d8b47b5265a/src/themes/oui-next/oui_next_colors_dark.scss#L41-L45

But we can definitely revisit or redefine these relationships.

src/global_styling/mixins/_form.scss
```diff
@@ -262,16 +262,21 @@
...
     // sass-lint:disable-block no-vendor-prefixes
-    &:-webkit-autofill {
-      -webkit-text-fill-color: lightOrDarkTheme($ouiColorDarkestShade, $ouiColorLightShade);
+    &:-webkit-autofill,
+    &:autofill {
+      /* Many browsers use `!important` in their built-in `:-webkit-autofill` style declarations,
+       * making their colors non-overridable. `-webkit-text-fill-color` is able to overwrite the
+       * appearance of texts and is used here as a workaround.
+       */
+      -webkit-text-fill-color: $ouiColorDarkestShade;
```
joshuarrrr, line 271: Why are we using the shade definition directly, and not `$ouiTextColor`?

joshuarrrr (commented):
Shouldn't all these styles be linked to `$ouiTextColor` rather than a shade var?

KrooshalUX:
@joshuarrrr  really helpful context, thank you for outlining the text values. I looked through my original definitions and it appears that the value from "we explicitly override $ouiTextColor and $ouiTitleColor definitions to use #DFE5EF" was perhaps a generated value - or did it get specified by engineering? 

Either way, I think there is opportunity for us to clean up these relationships and improve the documentation. If we have various permutations floating around of text color, this could be difficult to manage and adjust for accessibility.

joshuarrrr:
> I looked through my original definitions and it appears that the value from "we explicitly override $ouiTextColor and $ouiTitleColor definitions to use #DFE5EF" was perhaps a generated value - or did it get specified by engineering?

I'd have to trace back through the blame, but I think neither - I think it was set that way in the existing dark theme and we just copied it over as part of the creation of the Next theme.

> Either way, I think there is opportunity for us to clean up these relationships and improve the documentation. If we have various permutations floating around of text color, this could be difficult to manage and adjust for accessibility.

Agreed - specifically with a case like this, I think it's fair to say that any text color value should at least point at a text color variable, so that the definition is centralized.

src/global_styling/mixins/_form.scss
AMoo-Miki: `ouiColorDarkestShade` is what the current stylesheet uses and the value of `ouiTextColor` doesn't match `ouiColorDarkestShade` in dark themes.

BSFishy (approved):

joshuarrrr (dismissed):
@AMoo-Miki Re-reviewing the discussion here, I think we definitely want to use a text color variable for these values rather than a particular shade for consistency.

src/global_styling/mixins/_form.scss
joshuarrrr: I intended this comment for the other file 😁

joshuarrrr:
@AMoo-Miki I updated the values only in the next theme.

joshuarrrr (approved):

PR #879, opened 2023-07-12
Bump typescript to 4.6.4

### LINKED ISSUE #887: useUnknownInCatchVariables reset
Please reset `useUnknownInCatchVariables` to its default value to restore common behavior. 

Note that the `useUnknownInCatchVariables` was disabled in #879

The proposed change:
* Remove `useUnknownInCatchVariables` from tsconfig file

### LINKED ISSUE #888: Bump typescript to 4.6.4
Embracing the latest technologies is crucial for the long-term success of oui. I am eager to explore opportunities to modernize the underlying technology stack. We need to upgrade dependencies to the latest versions. It will reduce security risks and improve developer experience. 

This issue proposes step in this direction by upgrading typescript to version 4.6.4

gulderov:
### Description
Bump typescript to 4.6.4

### Issues Resolved

closes #888
closes #887 

### Check List
- [x] All tests pass
  - [x] `yarn lint`
  - [x] `yarn test-unit`


src/components/header/header_links/header_links.tsx
```diff
@@ -95,16 +95,15 @@ export const OuiHeaderLinks: FunctionComponent<OuiHeaderLinksProps> = ({
   popoverProps,
   ...rest
 }) => {
-  const { onClick: _onClick, iconType = 'apps', ...popoverButtonRest } = {
-    ...popoverButtonProps,
-  };
+  const { onClick, iconType = 'apps', ...popoverButtonRest } =
+    popoverButtonProps || {};
```
BSFishy, line 99: Can we use the nullish coalescing operator here instead of logical or?

src/components/list_group/pinnable_list_group/pinnable_list_group.tsx
```diff
@@ -115,7 +115,7 @@ export const OuiPinnableListGroup: FunctionComponent<OuiPinnableListGroupProps>
       );
 
       // Add the pinning action unless the item has it's own extra action
-      if (onPinClick && !itemProps.extraAction && pinnable) {
```
BSFishy, line 118: Why is `onPinClick` removed here? From a brief look at the code, it seems like it should be null checked

tsconfig.json
```diff
@@ -40,6 +40,9 @@
     // Disallow inconsistently-cased references to the same file.
     "forceConsistentCasingInFileNames": true,
 
+    // Keep type of the variable in a catch clause as `any`
+    "useUnknownInCatchVariables": false,
```
BSFishy, line 44: Why do we want to keep it as any? Is there a significant downside of using `unknown`? Will it involve a substantial amount of extra work? Could it be done in a followup task?

src-docs/src/components/guide_section/guide_section.tsx
```diff
@@ -135,7 +135,7 @@ export const GuideSection: FunctionComponent<GuideSection> = ({
     const isPlaygroundUnsupported =
       typeof window !== 'undefined' &&
       typeof document !== 'undefined' &&
-      !!window.MSInputMethodContext &&
+      !!(window as any).MSInputMethodContext &&
```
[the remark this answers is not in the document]
gulderov: Maybe, we could consider completely removing this. Please look at the follow up issue #886

src/components/list_group/pinnable_list_group/pinnable_list_group.tsx
gulderov: The `onPinClick` is defined as required prop for `OuiPinnableListGroup` component. Typescript will give error, if `onPinClick` is not provided. So no extra validation needed. I hope we can benefit from type checking here

tsconfig.json
gulderov: Typescript can enhance code quality, but it is also important to keep balance between type safety and ease of use

The `useUnknownInCatchVariables` option is useful because it enforces a higher level of type safety in `catch` blocks. However, enabling this option can also lead to additional complexity in code. We would need to add extra type checking code in each `catch` block.

Let me show on examples

It is common to expect catch variable as instance of Error:
``` typescript
try {
  throw Error('error message');
} catch (error) {
  console.log(error.message); // => LOG: 'error message'
}
```

but it is not always the case:
``` typescript
try {
  throw 'error message';
} catch (error) {
  console.log(error.message); // => LOG: undefined
}
```

`useUnknownInCatchVariables` is intended to catch errors like this:
``` typescript
try {
  throw 123;
} catch (error) {
  if (err instanceof Error) {
    console.log(err.message);
  } else if (typeof err === 'string') {
    console.log(err);
  } else {
    console.log('Unexpected error')

  }
}
```

or:
``` typescript
try {
  throw 123;
} catch (error) {
  const e = error instanceof Error ? error : Error( `Unexpected Error: ${error}`);
   console.log(e.message);
}
```

Same time it is much easier to avoid checks by ignoring types:
``` typescript
try {
    throw new OwnError(123);
} catch (error: unknown) {
    console.log((error as OwnError).message);
}
```

or even: 
``` typescript
try {
  throw new Error(123);
} catch (error : any) {
   console.log(error.message);
}
```


While enabling `useUnknownInCatchVariables` could potentially catch errors, it might also introduce extra complexity into codebase. I am cautious about making this change, especially since the `useUnknownInCatchVariables` is enabled by default.

 I would be happy to make this change, if benefits of enabling `useUnknownInCatchVariables` outweigh potential downsides. Please let me know if you want to proceed with this and I will create a followup task

package.json
```diff
@@ -159,8 +159,8 @@
     "@types/react-router-dom": "^5.3.3",
     "@types/url-parse": "^1.4.8",
     "@types/uuid": "^9.0.1",
-    "@typescript-eslint/eslint-plugin": "^4.8.1",
-    "@typescript-eslint/parser": "^4.8.1",
+    "@typescript-eslint/eslint-plugin": "^5.62.0",
```
gulderov, line 162: The `@typescript-eslint/eslint-plugin: "^6.0.0"` requires at least node v16, but `.nvmrc` contains `14.20.0`

src/components/header/header_links/header_links.tsx
SergeyMyssak: Since this prop can be either only undefined or only an object, it would be better to use `??`. But it looks good with `||` as well

src/components/suggest/suggest_item.tsx
```diff
@@ -193,9 +193,9 @@ export const OuiSuggestItem: FunctionComponent<OuiSuggestItemProps> = ({
   if (onClick) {
     return (
       <button
-        onClick={onClick}
         className={classes}
-        {...(rest as PropsForButton)}>
+        {...(rest as PropsForButton)}
+        onClick={onClick}>
```
SergeyMyssak, line 198: Are you sure we should put `onClick` after `{{...(rest as PropsForButton)}`? I understand that TS gives an error there, but it would still be good to understand why it started giving it

tsconfig.json
SergeyMyssak: It's a good idea to reduce complexity and remove conditions inside the catch block, but I think type safety and robust error handling have a higher weight than reducing complexity like that, so for me, it's better to use `unknown` rather than `any` for error variables in catch clauses

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### 💥 Breaking Changes
 
+- Upgraded to TypeScript v4.6.4
```
SergeyMyssak, line 7: No public interfaces seem to have been changed with the TypeScript version update, I don't think it's a breaking change

src/components/suggest/suggest_item.tsx
gulderov: Typescript highlights that `onClick` is specified more than once, so usage will be overwritten

This is due usage of `as PropsForButton`, which reintroduces the `onClick` while it is already picked
```type PropsForButton = Omit<
  ButtonHTMLAttributes<HTMLButtonElement>,
  'onClick' | 'type'
> & {
  onClick: MouseEventHandler<HTMLButtonElement> | undefined;
};
```

Provided solution is easer than `{...(rest as Omit<PropsForButton, 'onClick'}`or refactoring of the `PropsForButton`

src-docs/src/components/guide_section/guide_section.tsx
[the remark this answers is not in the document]
gulderov: Good point about fixing it by adding to global types. The usage of `any` is common to highlight areas of uncertainty or complexity in typescript. It is usually recommended to avoid it if possible

In this case it should be safe to mark some pain point by `as any`. It could be more scary to cleanup `index.d.ts` in future, than cleanup local usage of `any`. My preference is to keep the`index.d.ts` nice and clean. 

On other hand I cant see better cast here and we have issue #886 for refactoring

tsconfig.json
gulderov: There seems to be some confusion regarding this change. I have created issue #887 to do refactoring with `useUnknownInCatchVariables ` in separate task. It will help to keep this pull request limited. 

Could you please review it @BSFishy  @SergeyMyssak ?

SergeyMyssak: I'm wondering if we can avoid this change? I see that you have created an issue to remove it

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## [Unreleased](https://github.com/opensearch-project/oui/tree/main)
 
+- Bump TypeScript to v4.6.4 ([#879](https://github.com/opensearch-project/oui/pull/879))
+- Bump @typescript-eslint/eslint-plugin to v5.62.0 ([#879](https://github.com/opensearch-project/oui/pull/879))
+- Bump @typescript-eslint/parser to  v5.62.0 ([#879](https://github.com/opensearch-project/oui/pull/879))
+
```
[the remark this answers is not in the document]
gulderov: I have moved the changes to the Maintenance sections

Could you kindly review the corresponding issue #888 ?

tsconfig.json
gulderov: `useUnknownInCatchVariables` is not in tsconfig anymore. We have changes in this pull request to close #887

src/components/suggest/suggest_item.tsx
gulderov: Typing is chaged to suggested

src/components/header/header_links/header_links.tsx
gulderov: Fixed by `||` => `??`

SergeyMyssak (approved):
@gulderov well done, thank you for your contribution! All that's left is to fix the merge conflicts.

@BSFishy, are we going to backport this PR to `1.x`?

BSFishy (approved):

BSFishy:
> @BSFishy, are we going to backport this PR to `1.x`?

Yep, I don't see any public API changes, so this should be good to go.

I'd also like to get another pair of eyes or two on this, as its blast radius might be a little bit larger than we usually have. Pinging people now

src/components/datagrid/data_grid_inmemory_renderer.tsx
```diff
@@ -56,11 +56,7 @@ export interface OuiDataGridInMemoryRendererProps {
 function noop() {}
 
 function getElementText(element: HTMLElement) {
-  return 'innerText' in element
-    ? element.innerText
-    : // (this line left here to satisfy Prettier since a ts-ignore is used on the next line)
-      // @ts-ignore TypeScript thinks element.innerText always exists, however it doesn't in jest/jsdom environment
-      element.textContent || undefined;
+  return 'innerText' in element ? element.innerText : element.textContent || '';
```
joshuarrrr, line 59: It looks like we're changing the return type to `string` from `string | undefined` - is this intentional, and does it have any effect in caller contexts?

joshuarrrr (approved):
LGTM

src/components/datagrid/data_grid_inmemory_renderer.tsx
AMoo-Miki: While i think it will impact the caller contexts, i think the previous behavior of returning undefined was wrong.

src/components/list_group/pinnable_list_group/pinnable_list_group.tsx
AMoo-Miki: Compile-time warning cannot take the place of proper guarding. There are many scenarios that TS will not be able check during compiling.

That said, `onPinClick` check shouldn't be up there but should be added down.
```
if (onPinClick) itemProps.extraAction.onClick = () => onPinClick(item);
```

src/components/search_bar/query/default_syntax.ts
```diff
@@ -353,8 +353,10 @@ const validateFieldValue = (
     try {
       schemaField.validate(value);
     } catch (e) {
+      const message = e instanceof Error ? e.message : String(e);
+
       error(
-        `Invalid value \`${expression}\` set for field \`${field}\` - ${e.message}`,
+        `Invalid value \`${expression}\` set for field \`${field}\` - ${message}`,
         location
       );
```
AMoo-Miki, line 361: Let's strengthen this a bit:
```
     const message = (e instanceof Error ? e.message || e.code : String(e)).trim();

      error(
        `Invalid value \`${expression}\` set for field \`${field}\`${message ? ` - ${message}` : '' }`,
        location
      );
```

CHANGELOG.md
```diff
@@ -74,6 +74,7 @@
 
 ### 🛠 Maintenance
 
+- Bump TypeScript to v4.6.4 ([#879](https://github.com/opensearch-project/oui/pull/879))
```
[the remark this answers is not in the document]
AMoo-Miki: I moved it.

src/components/search_bar/query/default_syntax.ts
AMoo-Miki: I committed a nicer version of the above suggestion.

src/components/list_group/pinnable_list_group/pinnable_list_group.tsx
AMoo-Miki: I committed the above suggestion.

joshuarrrr:
Thanks @gulderov ! This was a very valuable contribution!

Issue #119, opened 2022-11-10
[PROPOSAL] Replace sass-lint with stylelint


BSFishy:
**Is your feature request related to a problem? Please describe.**

`sass-lint` is the linter we currently use for our SCSS files. It is very old, and as of writing this issue, the last release was published 4 years ago. Because of its age, sometimes issues arise, for example [referencing modules that have been removed from Webpack for 2 major versions](https://github.com/sasstools/sass-lint/issues/1324).

**Describe the solution you'd like**

A complete migration from sass-lint to Stylelint. As of writing, the last Stylelint release was published 7 days ago, and the project is significantly more active and it supports many more modern features. By replacing it, we will have more confidence that out linter won't break because it is unmaintained.

ashwin-pc:
We have replace the same in OSD a while ago with style lint. Ref: https://github.com/opensearch-project/OpenSearch-Dashboards/issues/551

I think style lint is a valid alternative and is already in use in OSD. You can also refer to the PR's there if you face any issue with the migration

anirudha:
Closing — this issue has been open for 1-3 years without progress. If still relevant, please reopen or create a new issue with updated context.

PR #942, opened 2023-07-31
Migrate to Dart Sass

BSFishy:
### Description
Migrate to Dart Sass instead of using `node-sass`.

A few things:
 - Replace `sass-extract` with `sass-extract-dart` to work with Dart Sass
    - Call `sass_extract`, as Node bindings are a little finicky and async doesn't seem to actually be running in async
 - Fix usage of slash division with `math.div`, as slash division is deprecated in Dart Sass
 - Use custom `exec` function in compile script to allow for proper async (Javascript eagerly executes promises, so we can't use `execSync` or wrap it in a custom promise)
 - Compile Sass and run sass-extract in parallel for easy speedup
 - Add array type to generated JSON typedef (`sass-extract-dart` supports extracting array types, which `sass-extract` didn't)

Questions:
 - This fails the linter, as in some places we use `#{math.div(...)}` syntax. `sass-lint` isn't able to parse this, where it throws an error. We could solve this in 2 ways: quick and easy or long and maintainable
    - Quick and easy: Split the usage into 2 lines: declare a variable with the value and use the variable. This is quick and easy to implement, but we should plan on using Stylelint at some point.
    - Long and maintainable: migrate to Stylelint. `sass-lint` in unmaintained and is clearly causing us issues. However, I don't know how much effort it might be to migrate. To compare with the migration that happened in OSD (https://github.com/opensearch-project/OpenSearch-Dashboards/pull/1413), 240 files changed, but it looks like most of them were fixing linter issues?
 - There are some differences in the generated JSON. Mainly it's supporting arrays and some differences in number handling. This _might_ be breaking, but I think all of these are affected outside of what we should consider public API (https://oui.opensearch.org/1.2/#/guidelines/sass). How should we handle this? Will get a diff soon

### Check List
- [ ] All tests pass
  - [ ] `yarn lint`
  - [ ] `yarn test-unit`


scripts/compile-scss.js
```diff
@@ -34,6 +34,17 @@ const postcssConfigurationWithMinification = {
...
 };
 
+function exec(command, options) {
+  const child = childProcess.exec(command, options);
+
+  return new Promise((resolve, reject) => {
+    child.addListener('error', reject);
+    child.addListener('exit', (...args) => {
+      resolve(...args);
+    });
```
BSFishy, line 44: Need to rewrite as

```suggestion
    child.addListener('exit', resolve);
```

scripts/compile-scss.js
```diff
@@ -120,16 +131,28 @@ async function compileScssFile(
...
+  const extractedVars = JSON.parse(
+    fs.readFileSync(outputVarsFilename, {
+      encoding: 'utf-8',
+    })
   );
 
+  // const result = await sass.compileAsync(inputFilename);
+  fs.writeFileSync(outputCssFilename, result.css);
+
+  const renderedCss = fs.readFileSync(outputCssFilename, { encoding: 'utf-8' });
```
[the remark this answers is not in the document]
BSFishy: unnecessary i/o

src/components/form/range/_range_highlight.scss
```diff
@@ -9,11 +9,13 @@
...
  */
 
+@use 'sass:math';
+
 .ouiRangeHighlight {
   position: absolute;
   left: 0;
   width: 100%;
-  top: calc(50% - #{($ouiRangeHighlightHeight / 2)});
+  top: calc(50% - #{math.div($ouiRangeHighlightHeight, 2)});
```
BSFishy, line 18: This is one of the cases where `sass-lint` fails to parse this syntax, for reference

joshuarrrr:
@BSFishy I see you generally went with the `math.div(...)` syntax. Can we avoid the linting errors for now by using `calc( foo / bar)` syntax instead? https://sass-lang.com/documentation/breaking-changes/slash-div/

SergeyMyssak (commented):
@BSFishy, really good changes! 
1) The solution provided by @joshuarrrr is working, we can use `calc` instead of `math.div` to avoid linter errors
2) We can indeed migrate to Stylelint, mostly the changes will be regarding automatic fixing by linter
3) I don't think that this is breaking change, but it would be good to take a look at diff

gulderov:
@BSFishy we have proposal #119 for migration to Stylelint. It seems reasonable to separate the migration into two distinct pull requests: one for Dart Sass and the other for Stylelint

gulderov:
@BSFishy we have `stylelint` in pull request #949 to support migration

BSFishy:
The more I work on this, the more I realize we will most likely need to profile our sass build to see where we're getting performance hits. I have a feeling it's the way we include files. Either way, I think I'm going to need to put this PR on hold until that gets fixed

BSFishy:
Superseded by #1001

PR #949, opened 2023-08-03
Replace `sass-lint` with `stylelint`

gulderov:
### Check List
- [ ] All tests pass
  - [ ] `yarn lint`
  - [ ] `yarn test-unit`


.stylelintrc.yml
```diff
@@ -0,0 +1,19 @@
+extends:
+  - stylelint-config-standard-scss
+rules:
+  # while we still use node-sass, only legacy rgb() notation is allowed
+  color-function-notation: "legacy"
+  # recommended to turn off descending specificity since we use a lot of nesting:
+  # https://stylelint.io/user-guide/rules/list/no-descending-specificity/
+  no-descending-specificity: null
```
BSFishy, line 8: It might be nice to have a selector ordering rule. We should look into this further to see how much work it would be and if we should do it in the first place

src-docs/src/components/guide_components.scss
```diff
@@ -177,7 +179,7 @@ $elasticLogoTextDark: #1C1E23;
 }
 
 .guideDemo__ghostBackground {
-  @if (lightness($ouiTextColor) < 50) {
+  @if  lightness($ouiTextColor) < 50 {
```
BSFishy, line 182: Why the extra space here? There seem to be a lot of these? Is it just an artifact of `yarn lint-sass-fix`?

```suggestion
  @if lightness($ouiTextColor) < 50 {
```

src-docs/src/theme_dark.scss
```diff
@@ -10,9 +10,8 @@
  */
 
 // sass-lint:disable no-url-domains, no-url-protocols
```
BSFishy, line 12: Would love to get rid of old `sass-lint` disables

BSFishy (commented):
Looks pretty good so far, just a few comments I had on first inspection

src-docs/src/theme_dark.scss
gulderov: `sass-lint` have been removed from the codebase

src-docs/src/components/guide_components.scss
gulderov: It is artifact from `yarn lint-style --fix`. I was wondering to fix it with `prettier`. But we need to update `prettier` first and it is another task.
 
I tend to avoid manual intervention and keep it as is for now. Prettier would reformat it later

.stylelintrc.yml
gulderov: We could handle it as separate task

BSFishy:
https://github.com/opensearch-project/oui/actions/runs/5759821731/job/15628275259?pr=949#step:5:13

Do you know why the linter is failing in CI? Does it need to be rerun?

gulderov:
@BSFishy I removed the quotes to try and fix CI

BSFishy:
Love to see the CI is passing, gives me a lot more confidence in this. Will re-review and approve if everything looks good

package.json
```diff
@@ -20,12 +20,12 @@
...
     "extract-i18n-strings": "node ./scripts/babel/fetch-i18n-strings",
-    "lint": "yarn tsc --noEmit && yarn lint-es && yarn lint-sass",
+    "lint": "yarn tsc --noEmit && yarn lint-es && yarn lint-style",
     "lint-fix": "yarn lint-es-fix",
     "lint-es": "eslint --cache \"{src,src-docs}/**/*.{ts,tsx,js}\" --max-warnings 0",
     "lint-es-fix": "yarn lint-es --fix",
-    "lint-sass": "sass-lint -v --max-warnings 0",
-    "lint-sass-fix": "sass-lint-auto-fix -c ./.sass-lint-fix.yml",
+    "lint-style": "stylelint \"{src,src-docs}/**/*.scss\" --max-warnings 0",
+    "lint-style-fix": "yarn lint-style --fix",
```
BSFishy, line 28: Curious if this will cause anything unintended to break. All CI passes, but I wonder if anyone else is using these commands that we don't know about. Either way, I don't think it's something we _really_ need to worry about for breaking things ~ this is more meant for dev anyway

package.json
```diff
@@ -216,6 +210,7 @@
     "moment-timezone": "^0.5.41",
     "node-sass": "npm:@amoo-miki/node-sass@9.0.0-libsass-3.6.5",
     "pegjs": "^0.10.0",
+    "postcss": "^8.4.28",
```
BSFishy, line 213: I am astonished that we were using `postcss` without specifying it as a dep 🤦

BSFishy (commented):
Overall looks good, will pull down and play around with it then approve

.stylelintignore
```diff
@@ -0,0 +1,2 @@
+build
+target
```
BSFishy, line 2: We have a few artifact directories:

```
dist
docs
es
lib
test-env
types
```

Additionally, would `node_modules` be a good choice to be here?

.stylelintrc.yml
```diff
@@ -0,0 +1,19 @@
+extends:
+  - stylelint-config-standard-scss
+rules:
+  # while we still use node-sass, only legacy rgb() notation is allowed
+  color-function-notation: "legacy"
```
BSFishy, line 5: Will need to followup on this for #1001

BSFishy (commented):
Just a few minor final things

Issue #940, opened 2023-07-31
Increase contrast of keyboard focus highlight when applied to OuiLink


KrooshalUX:
**Current theme has stronger perceptual contrast between link text and keyboard highlight:** 

<img width="406" alt="Screen Shot 2023-08-07 at 1 39 06 PM" src="https://github.com/opensearch-project/oui/assets/105819041/5e92432f-6119-4d9e-a3a3-eccaa944d2c2">
<img width="418" alt="Screen Shot 2023-08-07 at 1 39 51 PM" src="https://github.com/opensearch-project/oui/assets/105819041/876a9639-2ed4-4373-ab38-d04586c686a9">

**Next theme:**
<img width="414" alt="Screen Shot 2023-08-07 at 1 40 23 PM" src="https://github.com/opensearch-project/oui/assets/105819041/fe49eb66-1521-494c-92ed-386f98cb094f">
<img width="399" alt="Screen Shot 2023-08-07 at 1 40 49 PM" src="https://github.com/opensearch-project/oui/assets/105819041/3e2a0ebb-0afc-4962-a8eb-fd5d769db88f">

While next theme is passing a11y contrast test in my browser, it doesn't seem that legible to me. @joshuarrrr  can you help me understand the shading/tinting that happens with the primary color as it applies to link and highlight?

KrooshalUX:
@joshuarrrr  updated with screenshots.

PR #962, opened 2023-08-14
Clean up focus background implementation

### LINKED ISSUE #948: Increase contrast of keyboard focus highlight when applied to OuiLink

BSFishy:
### Description
Clean up some of the focus background implementation details

 * Make `ouiFocusBackground` mixin actually use `$color` argument
 * Make `$ouiFocusTransparency` an actual percentage instead of a float [0, 1]
 * Refactor any usage of `$ouiFocusTransparency` to respect it as a percentage instead of a float [0, 1]

#### Before

Screenshot (134)
Screenshot (135)
Screenshot (136)
Screenshot (137)
Screenshot (138)
Screenshot (139)
Screenshot (140)
Screenshot (141)

#### After

Screenshot (142)
Screenshot (143)
Screenshot (144)
Screenshot (145)
Screenshot (146)
Screenshot (147)
Screenshot (148)
Screenshot (149)

### Related issues

Fixes #948

### Check List
- [x] All tests pass
  - [x] `yarn lint`
  - [x] `yarn test-unit`


src/global_styling/variables/_states.scss
```diff
@@ -20,8 +20,8 @@ $ouiFocusRingSizeLarge: $ouiSizeXS !default;
 $ouiFocusRingSize: $ouiFocusRingSizeLarge * .75 !default;
 
 // Transparency
-$ouiFocusTransparency: lightOrDarkTheme(.1, .3) !default;
-$ouiFocusBackgroundColor: tintOrShade($ouiColorPrimary, ((1 - $ouiFocusTransparency) * 100%), ((1 - $ouiFocusTransparency) * 100%)) !default;
+$ouiFocusTransparency: lightOrDarkTheme(90%, 70%) !default;
```
joshuarrrr, line 23: This naming is still misleading. How about something like `$ouiFocusChangePercent`? Because it's ultimately passed to either `shade` or `tint` functions as the `$percent` param.

src/global_styling/variables/_states.scss
```diff
@@ -20,8 +20,8 @@ $ouiFocusRingSizeLarge: $ouiSizeXS !default;
...
+$ouiFocusBackgroundColor: tintOrShade($ouiColorPrimary, $ouiFocusTransparency, $ouiFocusTransparency) !default;
```
joshuarrrr, line 24: The simplification reveals that this definition itself is a bit suspect. If you search the OUI stylesheets for `tintOrShade` usage, you'll see that it's most common to specify more `tint` than `shade` (or, occasionally, the reverse). That makes sense from a color perception perspective.

There are only 2 other examples I found where the two percentages are the same:

https://github.com/opensearch-project/oui/blob/fe979c3662afc5e4b620f05c838d4b98beb7573e/src/global_styling/variables/_buttons.scss#L16

https://github.com/opensearch-project/oui/blob/fe979c3662afc5e4b620f05c838d4b98beb7573e/src/components/control_bar/_control_bar.scss#L177

So I don't think we should be afraid to further update this definition to be more in tune with other derived colors.

joshuarrrr (commented):
looks good, but I'd like to get rid of "Transparency" in the var name.

Also need a changelog entry

src/global_styling/variables/_states.scss
BSFishy: I think the point of tint or shade is more to specifically get the tint vs the shade functionality depending on dark or light colors. The actual color switching is done by the `$ouiFocusTransparency` variable, so we will only ever get tint 90% on light mode and shade 70% on dark mode. It's just separated into 2 separate variables

BSFishy:
I changed the shade amount to 65%. It _seems_ to be a little better but I cant check with next theme because some of the colors are detected as dark when they should be light and vice versa

joshuarrrr (approved):

AMoo-Miki (approved):

KrooshalUX:
Does this address https://github.com/opensearch-project/oui/issues/940 ? If so, LGTM!!!

BSFishy:
Yep, also changes contrast :)

PR #1017, opened 2023-09-05
Add Dashboards integration workflow

BSFishy:
### Description
Adds a workflow to run tests in Dashboards with current OUI changes. This is meant to establish a pattern for how integrations could be run with Dashboards.

Also did some cleanup:
 - Split build and build docs jobs, so they can run in parallel
 - Put `name` at the top of all jobs to keep everything the same

### Check List
- [x] All tests pass
  - [x] `yarn lint`
  - [x] `yarn test-unit`


AMoo-Miki (approved):

.github/workflows/build_and_test.yml
```diff
@@ -10,10 +10,15 @@ on:
     paths-ignore:
       - '**/*.md'
 
+env:
+  CI: 1
+  GCS_UPLOAD_PREFIX: fake
```
joshuarrrr, line 15: curiosity - what does this do?

joshuarrrr (approved):
This is awesome! Hopefully we don't add too much time or false positives to our OUI PRs, but I think this should be a big win.

.github/workflows/build_and_test.yml
BSFishy: Not completely sure. When I ran it without this var, one of the tests failed, so I just copied it from the OSD build and test workflow

BSFishy:
> This is awesome! Hopefully we don't add too much time or false positives to our OUI PRs, but I think this should be a big win.

Exactly. This should ideally push down the number of false positives we have, as it adds an automated check to see if OUI integrated into OSD effectively (i.e. we're not breaking anything), so we can catch issues like that when the PR is made, rather than 3 days before code freeze.

PR #1024, opened 2023-09-07
Add exit code to compile-scss script on failure

BSFishy:
### Description
Add exit code to compile-scss script on failure. Previously, if this script had an error, it wouldn't propagate, which would cause false positives in our CI workflows.

### Check List
- [x] All tests pass
  - [x] `yarn lint`
  - [x] `yarn test-unit`

### DIFF
--- a/scripts/compile-scss.js
+++ b/scripts/compile-scss.js
@@ -74,6 +74,9 @@ async function compileScssFiles(
             error.stack
           }`
         );
+
+        process.exitCode = 1;
+        return;
       }
 
 
@@ -99,6 +102,8 @@ async function compileScssFiles(
             error.stack
           }`
         );
+
+        process.exitCode = 1;
       }
       /* End of Aliases */
     })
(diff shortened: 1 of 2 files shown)


scripts/compile-scss.js
```diff
@@ -74,6 +74,9 @@ async function compileScssFiles(
             error.stack
           }`
         );
+
+        process.exitCode = 1;
+        return;
```
BSFishy, line 79: This return is a failfast mechanism to prevent the EUI aliased compilation from running and spitting out more errors

joshuarrrr (approved):
Do you want these to be the same exit code, or different to distinguish between them?

BSFishy:
> Do you want these to be the same exit code, or different to distinguish between them?

I don't think it matters much for this case, as long as it isn't 0, the workflows will fail. Maybe it could be useful for some case, but the point of this PR is to make the command propagate errors. If we find a use case in the future, this is an easy thing to change

AMoo-Miki (approved):

PR #1028, opened 2023-09-08
Update docsite version selector to use OUI

BSFishy:
### Description
Previously, the version selector on the docsite was a custom implementation that was built on Web Components. This implementation is roughly equivalent to the previous one, except it uses OUI and should have simpler logic (i.e. not pinging `/versions.json` on a local instance)

Screenshot (149)

### Check List
- [x] All tests pass
  - [x] `yarn lint`
  - [x] `yarn test-unit`


BSFishy:
> Is there no OUI component that we can style for this? If no, is there any value in introducing this as a component to OUI?

I was trying to use super select, but there are some oddities with its width calculation that make it not a feasible option for this case. The theme selector is implemented with this pattern, so I used this one instead

src-docs/src/components/guide_version_selector/guide_version_selector.tsx
```diff
@@ -0,0 +1,113 @@
...
+  selected,
+}) => {
+  const isLocalDev = window.location.host.includes('803');
+  const [isPopoverOpen, setPopoverOpen] = useState(false);
+  const [selectedOption, setSelectedOption] = useState<string>(selected);
+  const [options, setOptions] = useState<string[]>([selected]);
+
+  const onButtonClick = () => {
+    setPopoverOpen(!isPopoverOpen);
+  };
```
SergeyMyssak, line 33: We can write it this way:
```tsx
const togglePopover = () => setPopoverOpen((prevState) => !prevState);
```

SergeyMyssak (approved):
Looks good to me!

andreymyssak (approved):

src-docs/src/components/guide_version_selector/guide_version_selector.tsx
```diff
@@ -0,0 +1,113 @@
...
+      e.preventDefault();
+
+      closePopover();
+      setSelectedOption(value);
+      window.location.href = `/${value}`;
+    },
+    []
+  );
+
+  useEffect(() => {
```
joshuarrrr, line 50: nit - does this need to be an effect? The only dep is `isLocalDev`, which we don't expect to change - it's just a static property of the endpoint.

joshuarrrr (approved):
A couple questions, but looks good!

src-docs/src/components/guide_version_selector/guide_version_selector.tsx
```diff
@@ -0,0 +1,113 @@
...
+import { OuiPopover } from '../../../../src/components/popover';
+
+export type GuideVersionSelectorProps = {
+  selected: string;
+};
+
+export const GuideVersionSelector: FunctionComponent<GuideVersionSelectorProps> = ({
+  selected,
+}) => {
+  const isLocalDev = window.location.host.includes('803');
```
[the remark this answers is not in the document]
BSFishy: This was carried over from the previous implementation, but I like that solution better, will fix

src-docs/src/components/guide_version_selector/guide_version_selector.tsx
BSFishy: Yes! The `useEffect` is important because otherwise we'll be running a `fetch` every time the component is rerendered. Adding `isLocalDev` as a dep is mainly to satisfy the linter, but also a best practice, just in case ;)

PR #1058, opened 2023-10-04
Fix docgen babel plugin

BSFishy:
### Description

After merging the Typescript update, there was a bug that popped up when trying to start the docsite. Essentially, the docgen info wasn't being generated. This is because the `react-docgen-typescript` package used a Typescript API that apparently broke in one of the versions between what we updated from and to. Updating that package fixes the issue.

### Check List
- [x] All tests pass
  - [x] `yarn lint`
  - [x] `yarn test-unit`

### DIFF
--- a/scripts/compile-clean.js
+++ b/scripts/compile-clean.js
@@ -11,6 +11,7 @@
 
 const { rimraf } = require('rimraf');
 
+rimraf.sync('.cache-loader');
 rimraf.sync('dist');
 rimraf.sync('lib');
 rimraf.sync('es');
--- a/package.json
+++ b/package.json
@@ -222,7 +222,7 @@
     "puppeteer": "^19.11.1",
     "raw-loader": "^4.0.1",
     "react": "^16.14.0",
-    "react-docgen-typescript": "^1.22.0",
+    "react-docgen-typescript": "^2.2.2",
     "react-dom": "^16.12.0",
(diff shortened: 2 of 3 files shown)


BSFishy:
`skip-changelog` label, as this will go in the same release as the Typescript update will go into

scripts/compile-clean.js
```diff
@@ -11,6 +11,7 @@
 
 const { rimraf } = require('rimraf');
 
+rimraf.sync('.cache-loader');
```
BSFishy, line 14: I added this, as it is something I would expect the `yarn clean` command to do. I was confused as to why my main clone wasn't working but a clean clone was working, and this was the reason. Just makes it easier for us to start off from a clean slate.

AMoo-Miki (approved):

seanneumann (approved):
🎉

joshuarrrr (approved):

joshuarrrr:
@BSFishy Might want to look into the failing OSD tests - can't see why this particular change would be a problem, but we should at least understand the reason.

BSFishy:
> @BSFishy Might want to look into the failing OSD tests - can't see why this particular change would be a problem, but we should at least understand the reason.

These are issues on OSD side - `@types/react` different version. Potentially just a package.json resolution. Need to test more on OSD side (I'm pretty sure this was expected)

Issue #739, opened 2023-05-01
(OUI Docs) Add link to figma community 


KrooshalUX:
The UX team has published the OUI Component Library in Figma and has made it available to the community. To guide designers and contributors to the file, a link to the Figma community file should be placed in the OUI documentation site header. 

- [ ] Add Figma logo SVG to header as a link , to the right of the GitHub icon (asset : https://github.com/opensearch-project/oui/blob/main/src-docs/src/images/logo-figma.svg) 
- [ ] Tooltip on hover: Figma component library
- [ ] URL: https://www.figma.com/community/file/1240039515407642648

navinkrv:
i want to contribute to this issue please assign me

navinkrv:
sorry for this but please assign it to another person

MadaniKK:
Hi! I would like to contribute to this issue. Thank you!

PR #1064, opened 2023-10-06
Add figma link on navbar

MadaniKK:
### Description
I have added the Figma icon with the OUI Figma component library link in the nav bar.
I did not write any unit tests yet.

Thank you for reviewing my PR!

### Issues Resolved
Issue #739

### Check List
- [☑️] All tests pass
- [☑️] `yarn lint`
- [☑️] `yarn test-unit`
- [☑️] Commits are signed per the DCO using --signoff

### Website Preview
image
image


MadaniKK:
Hi, I am not sure where would I write in the CHANGELOG.md. Should I write in the 1.0 section?

BSFishy:
> Hi, I am not sure where would I write in the CHANGELOG.md. Should I write in the 1.0 section?

You should put it in the unreleased section :)

src-docs/src/components/guide_page/guide_page_header.tsx
```diff
@@ -62,6 +62,25 @@ export const GuidePageHeader: React.FunctionComponent<{}> = () => {
     );
   }
 
+  function renderFigma() {
+    const href = 'https://www.figma.com/community/file/1240039515407642648';
+    const label = 'OUI Figma component library';
+    return isMobileSize ? (
+      <OuiButtonEmpty size="s" flush="both" iconType="logoGithub" href={href}>
+        {label}
+      </OuiButtonEmpty>
```
BSFishy, line 71:
```suggestion
      <OuiButtonEmpty size="s" flush="both" iconType="logoFigma" href={href}>
        {label}
      </OuiButtonEmpty>
```

Should this be `logoFigma`?

src/components/icon/assets/logo_figma.svg
BSFishy, line 1: This file should have the copyright header

src/components/icon/assets/logo_figma.js
BSFishy, line 1: This file should have the copyright header

BSFishy, line 1: Also make sure this file is generated by `yarn compile-icons` instead of done manually :)

BSFishy (commented):
Good start. We'll need to discuss actually adding the Figma logo as an icon, specifically with colors built in. I know with other icons we don't specify colors so the builder can specify which color they want it.

You'll probably also need to run `yarn test-unit -u` to update snapshots for icon

src/components/icon/assets/logo_figma.js
MadaniKK: ah, I see... I was new to this and confused about how all these worked so I just did it manually :)...So does that mean now I should:
1. add the copyright header to logo_figma.svg (I copy-pasted the svg file from where the issue says it would be)
2. delete the figma.js file and generate a new one using yarn compile-icons
3. run yarn test-unit -u to update the snapshot
4. update CHANGELOG.md

Thank you!

BSFishy: Yeah, we generally follow this guide when adding icons: https://github.com/opensearch-project/oui/blob/main/wiki/creating-icons.md#add-the-icon-to-the-oui-repo

That goes through the entire process start to finish, but you have the general idea down

MadaniKK: oh wow got it. Thank you! I did not see there was a guide on this!!

MadaniKK: Done!

src-docs/src/components/guide_page/guide_page_header.tsx
```diff
@@ -62,6 +62,25 @@ export const GuidePageHeader: React.FunctionComponent<{}> = () => {
...
+    const label = 'OUI Figma component library';
+    return isMobileSize ? (
+      <OuiButtonEmpty size="s" flush="both" iconType="logoFigma" href={href}>
+        {label}
+      </OuiButtonEmpty>
+    ) : (
+      <OuiToolTip content="Figma Component Library">
+        <OuiHeaderSectionItemButton
+          aria-label={label}
+          // @ts-ignore TODO: FIX
```
joshuarrrr, line 76: what's the typescript error here?

joshuarrrr (commented):
Looks good! Can we resolve the @ts-ignore?

BSFishy:
Thinking about it more, I feel like it would be better to use a static asset for the Figma logo rather than adding it as an icon. Looping in @kgcreative for opinions

joshuarrrr:
> Thinking about it more, I feel like it would be better to use a static asset for the Figma logo rather than adding it as an icon. Looping in @kgcreative for opinions

It's funny, initially I was leaning that way, but seeing the PR convinced me this way is fine. Especially since we already have a bunch of other logos in there 🤷

BSFishy:
> It's funny, initially I was leaning that way, but seeing the PR convinced me this way is fine. Especially since we already have a bunch of other logos in there 🤷

I think the biggest thing for me is that the Figma logo has associated colors, which would mean that it would override the color prop of the icon. I think it just makes more sense as a static asset rather than an icon, since it has a specific use-case rather than being meant for more universal usage

kgcreative:
Makes sense to me. I'm aligned with the figma logo being a static asset

src-docs/src/components/guide_page/guide_page_header.tsx
MadaniKK: I actually don't know, I copy-pasted the renderGithub() and modified based on that. Now that I have looked at it, I think it is related to prettier formatting, already fixed it.

MadaniKK:
> > It's funny, initially I was leaning that way, but seeing the PR convinced me this way is fine. Especially since we already have a bunch of other logos in there 🤷
> 
> I think the biggest thing for me is that the Figma logo has associated colors, which would mean that it would override the color prop of the icon. I think it just makes more sense as a static asset rather than an icon, since it has a specific use-case rather than being meant for more universal usage

So what should I do now :)

BSFishy:
> So what should I do now :)

Yeah, should be pretty easy. You can remove all of the icon related stuff (I would recommend just going through the "Files changed" tab and finding what changed in relation to icons). Then, according to icon docs (https://oui.opensearch.org/1.3/#/display/icons#custom-svgs), you can use custom SVGs. So you'd import the svg something like this:

```ts
import figmaLogo from '<relative path to figma logo svg>';
```

The Figma logo seems to already exist in the repo here: https://github.com/opensearch-project/oui/blob/main/src-docs/src/images/logo-figma.svg

Then, to use it, you should just need to use a `OuiIcon` something like this:

```tsx
<OuiIcon type={figmaLogo} aria-hidden="true" />
```

(make sure to check it on a mobile size too to ensure that one is correct too :)

MadaniKK:
> > So what should I do now :)
> 
> Yeah, should be pretty easy. You can remove all of the icon related stuff (I would recommend just going through the "Files changed" tab and finding what changed in relation to icons). Then, according to icon docs (https://oui.opensearch.org/1.3/#/display/icons#custom-svgs), you can use custom SVGs. So you'd import the svg something like this:
> 
> ```ts
> import figmaLogo from '<relative path to figma logo svg>';
> ```
> 
> The Figma logo seems to already exist in the repo here: https://github.com/opensearch-project/oui/blob/main/src-docs/src/images/logo-figma.svg
> 
> Then, to use it, you should just need to use a `OuiIcon` something like this:
> 
> ```tsx
> <OuiIcon type={figmaLogo} aria-hidden="true" />
> ```
> 
> (make sure to check it on a mobile size too to ensure that one is correct too :)
> 
> Feel free to ping me here or on Slack if you get stuck!

Thank you for your instructions! I have made the changes.

BSFishy (approved):
LGTM

joshuarrrr (approved):

Issue #818, opened 2023-06-19
Define visualization color palettes


joshuarrrr:
As part of [providing a cohesive visualization experienc](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/2840)e, and to [provide updated look and feel](https://github.com/opensearch-project/OpenSearch-Dashboards/issues/4298), we need design system guidance on visualization color palettes. In the meantime, engineering is working on https://github.com/opensearch-project/OpenSearch-Dashboards/issues/4318 to prepare for these updates

The deliverables here can be in the form of updated OUI variables or simply documented guidance for color usage within visualizations.

KrooshalUX:
+ UX reminder task (for self) related to findings with [this maps pr](https://github.com/opensearch-project/OpenSearch-Dashboards/pull/4293): color update should address issues with lightest tints becoming imperceptible value

KrooshalUX:
This was being tracked here: https://github.com/opensearch-project/oui/issues/807 but I have linked the two together.

joshuarrrr:
@KrooshalUX @wbeckler My understanding is that we will still tweak these palettes at some point, but not for 2.10. Is there any targeted release for definition changes to visualization colors (either categorical or quantitative)? We've also discussed the possibility of dark/light-mode specific palettes, which would require some (minor) refactoring of the color utilities in OUI.

KrooshalUX:
This is the first I am hearing tweaked colors not going into 2.10. We are really close on the UI colors and I am wrapping up light mode changes (minimal) and now with the Visual Consistency Dashboard that went up last week, I have what I need to inform any of the changes (we are also updating the logic around repeated colors, correct?)

anirudha:
Closing — superseded by the V9 theme initiative (#1695) and OUI 2.0 work. The new theme system addresses the goals of this issue. See the V9 RFC for the current direction.

PR #1111, opened 2023-10-22
[Fix] "Guidelines" Documentation Render Blank Page

### LINKED ISSUE #1109: [BUG] "Guidelines" Documentation Render Blank Page
**Describe the bug**

When navigating to the Source Documentation page and clicking on either the "Colors" or "Sass" sections, both render blank pages instead of displaying the relevant documentation.

**To Reproduce**

1. On the side nav, click on either "Colors" or "Sass"
3. Observe that both pages render blank page

**Expected behavior**

Upon clicking on "Colors" or "Sass", users should be able to see the corresponding documentation for these sections instead of a blank page.

**Screenshots**

https://github.com/opensearch-project/oui/assets/65143821/6ab9b421-eeb1-4a55-bad4-46d4c4db3096

Willie-The-Lord:
### Description

Resolve the issue where the Source Documentation would render blank pages when users clicked on either the "Colors" or "Sass" sections.

### Issues Resolved

Closes #1109 

#### Before

https://github.com/opensearch-project/oui/assets/65143821/8f09f867-b4f0-411f-aa1c-b147810405f1

#### After

https://github.com/opensearch-project/oui/assets/65143821/8ecd4720-4751-4221-ba0e-228354ac9bb5

### Check List
- [x] All tests pass
  - [x] `yarn lint`
  - [x] `yarn test-unit`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
 
 - Add exit code to compile-scss script on failure ([#1024](https://github.com/opensearch-project/oui/pull/1024))
 - Correct file path for import of Query component ([#1069](https://github.com/opensearch-project/oui/pull/1069))
-- Resolve "Guidelines" Documentation Render Blank Page ([#1111](https://github.com/opensearch-project/oui/pull/1111))
+- Fix "Guidelines" documentation links rendering blank pages ([#1111](https://github.com/opensearch-project/oui/pull/1111))
 
 ### 🚞 Infrastructure
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 
 - Add exit code to compile-scss script on failure ([#1024](https://github.com/opensearch-project/oui/pull/1024))
 - Correct file path for import of Query component ([#1069](https://github.com/opensearch-project/oui/pull/1069))
+- Resolve "Guidelines" Documentation Render Blank Page ([#1111](https://github.com/opensearch-project/oui/pull/1111))
 
 ### 🚞 Infrastructure
 
--- a/src-docs/src/views/guidelines/colors/_utilities.js
+++ b/src-docs/src/views/guidelines/colors/_utilities.js
@@ -11,8 +11,9 @@
 
 import React, { useContext } from 'react';
 import { ThemeContext } from '../../../components';
-import { calculateContrast, rgbToHex } from '../../../../../src/services';
+import { calculateContrast } from '../../../../../src/services';
 import { getSassVars } from '../_get_sass_vars';
+import { hexToRgb } from '../../../../../src/services/color/hex_to_rgb';
 
 import { OuiBadge, OuiCopy, OuiFlexItem } from '../../../../../src/components';
 import { OuiIcon } from '../../../../../src/components/icon';
@@ -74,8 +75,8 @@ export const ratingAll = <OuiBadge color="#eee">ALL</OuiBadge>;
 
 function getContrastRatings(background, foreground, palette) {
   const contrast = calculateContrast(
-    [palette[background].r, palette[background].g, palette[background].b],
-    [palette[foreground].r, palette[foreground].g, palette[foreground].b]
+    hexToRgb(palette[background]),
+    hexToRgb(palette[foreground])
   );
 
   let contrastRating;
@@ -151,8 +152,8 @@ color: $${foreground};`;
             onClickAriaLabel="Click to copy SASS configurations"
             disabled={!contastIsAcceptableToCopy}
             style={{
-              backgroundColor: palette[background].rgba,
-              color: palette[foreground].rgba,
+              backgroundColor: palette[background],
+              color: palette[foreground],
             }}>
             {foreground}
           </OuiBadge>
@@ -161,8 +162,3 @@ color: $${foreground};`;
     </OuiFlexItem>
   );
 };
-
-export function getHexValueFromColorName(palette, colorName, key) {
-  const hex = key ? palette[colorName][key] : palette[colorName];
-  return rgbToHex(hex.rgba).toUpperCase();
-}
--- a/src-docs/src/views/guidelines/colors/vis_palette.js
+++ b/src-docs/src/views/guidelines/colors/vis_palette.js
@@ -21,9 +21,9 @@ import {
   OuiText,
   OuiPanel,
 } from '../../../../../src/components';
-import { rgbToHex } from '../../../../../src/services';
 
 export const VisPalette = ({ variant }) => {
+  // TODO: Make this dynamic once we have ouiPaletteColorBlind in the dark and next-dark theme
   const visColors = getSassVars('light').ouiPaletteColorBlind;
   const visColorKeys = Object.keys(getSassVars('light').ouiPaletteColorBlind);
 
@@ -41,7 +41,7 @@ export const VisPalette = ({ variant }) => {
                   onClick={copy}
                   size="xl"
                   type="stopFilled"
-                  color={rgbToHex(hex.rgba)}
+                  color={hex}
                 />
               )}
             </OuiCopy>
@@ -54,7 +54,7 @@ export const VisPalette = ({ variant }) => {
           <OuiFlexItem>
             <OuiText size="s" color="subdued">
               <p>
-                <code>{rgbToHex(hex.rgba).toUpperCase()}</code>
+                <code>{hex}</code>
               </p>
             </OuiText>
           </OuiFlexItem>
--- a/src-docs/src/views/guidelines/colors/vis_palette.js
+++ b/src-docs/src/views/guidelines/colors/vis_palette.js
@@ -23,7 +23,7 @@ import {
 } from '../../../../../src/components';
 
 export const VisPalette = ({ variant }) => {
-  // TODO: Make this dynamic once we have ouiPaletteColorBlind in the dark and next-dark theme
+  // ouiPaletteColorBlind() is currently shared across themes and dark/light modes. This will need to be  made dynamic once we have different visualization palettes as in https://github.com/opensearch-project/oui/issues/818
   const visColors = getSassVars('light').ouiPaletteColorBlind;
   const visColorKeys = Object.keys(getSassVars('light').ouiPaletteColorBlind);
 
--- a/src-docs/src/views/guidelines/sass.js
+++ b/src-docs/src/views/guidelines/sass.js
@@ -14,7 +14,6 @@ import sizes from '!!variables-from-scss!!../../../../src/global_styling/variabl
 import zindexs from '!!variables-from-scss!!../../../../src/global_styling/variables/_z_index.scss';
 import animations from '!!variables-from-scss!!../../../../src/global_styling/variables/_animations.scss';
 import breakpoints from '!!variables-from-scss!!../../../../src/global_styling/variables/_responsive.scss';
-import { rgbToHex } from '../../../../src/services';
 
 import { Link } from 'react-router-dom';
 
@@ -113,7 +112,7 @@ function renderPaletteColor(palette, color) {
       <OuiFlexItem grow={false}>
         <div
           className="guideSass__swatch"
-          style={{ background: rgbToHex(palette[color].rgba).toUpperCase() }}
+          style={{ background: palette[color] }}
         />
       </OuiFlexItem>
       <OuiFlexItem grow={false}>
--- a/src-docs/src/views/guidelines/colors/color_section.js
+++ b/src-docs/src/views/guidelines/colors/color_section.js
@@ -23,7 +23,6 @@ import {
   OuiPanel,
 } from '../../../../../src/components';
 import {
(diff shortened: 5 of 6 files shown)


Willie-The-Lord:
While investigating this issue, I found it challenging to identify the bug due to the implicit typing of variables. 

For example, the main issue lies in the fact that `palette[color]` is already a `string` in the format "#017D73". Therefore, the original code `rgbToHex(palette[color].rgba).toUpperCase()` is unnecessary and prone to errors.

Do we have any plans to migrate some of these files to TypeScript for better type safety?

src-docs/src/views/guidelines/colors/vis_palette.js
```diff
@@ -21,9 +21,9 @@ import {
   OuiText,
   OuiPanel,
 } from '../../../../../src/components';
-import { rgbToHex } from '../../../../../src/services';
 
 export const VisPalette = ({ variant }) => {
+  // TODO: Make this dynamic once we have ouiPaletteColorBlind in the dark and next-dark theme
```
BSFishy, line 26: If I'm not mistaken, this is intentionally not changed between themes and just provides Scss access to the variables generated by https://oui.opensearch.org/1.3/#/utilities/color-palettes

BSFishy: Looping in @kgcreative for more context

BSFishy (approved):

vvavdiya:
@Willie-The-Lord 
PR details need to update in CHANGELOG.md

vvavdiya (commented):
@Willie-The-Lord
PR details need to update in CHANGELOG.md

Willie-The-Lord:
> @Willie-The-Lord PR details need to update in CHANGELOG.md

Thank you for the heads up!

src-docs/src/views/guidelines/colors/_utilities.js
```diff
@@ -151,8 +152,8 @@ color: $${foreground};`;
             onClickAriaLabel="Click to copy SASS configurations"
             disabled={!contastIsAcceptableToCopy}
             style={{
-              backgroundColor: palette[background].rgba,
-              color: palette[foreground].rgba,
+              backgroundColor: palette[background],
+              color: palette[foreground],
```
BigSamu, line 156: @Willie-The-Lord, should not be here rgba values? Although I think is not a big deal, because the result is the same. I suppose that that property .rgba was a string in the format "rgba(<0-255>,<0-255>,<0-255>,<0-1>).

Just saying to keep consistency with what there was before. However, I don't see any utility function that returns the equivalent format for RGBA, just one for converting a HEX to RGB, but in array format.

BigSamu (approved):
Appart from the comment I did that is minor, the changes LGTM! I have already reviewed the code and tested it locally. Indeed the problem was tricky as @Willie-The-Lord mentioned.

BigSamu:
@BSFishy,

Additionally I want to comment the following when doing `console.log(palette)` for the aux function `getSassVars(theme)` in the file `src-docs/src/views/guidelines/_get_sass_vars.js`

Why do we repeat the colour themes using variables that have the suffix `Text`. For example, take a look at the screenshot below. We are repeating the values for hex colour`"#F66"` in the keys`euiDangerColor` and `euiDangerColorText`. In both cases, the value is a string. Shouldn't we avoid DRY?

image

src-docs/src/views/guidelines/colors/_utilities.js
Willie-The-Lord: Hi @BigSamu, I've just re-run the original code and `palette[background].rgba` is causing the render error as I mentioned in previous comments. This is because `palette[background]` returns a `string`, which does not have an `.rgba` property.

#### ❌ `palette[background].rgba`
Screenshot 2023-10-25 at 1 16 29 PM

#### ✅ `palette[background]`
Screenshot 2023-10-25 at 1 15 55 PM

BigSamu: I know! What I am trying to say is that the code was using RGBA format before and maybe for something (I don't know). That is why I suggest sticking with that format instead of HEX format. Make sense? However, you will need to build utility functions for getting RGBA values. It is just for consistency. 

Of course, It is a really minor issue. Your solution is much more simple and straightforward.

Willie-The-Lord: Ah, I understand now. I can also take on the task of building the utility function. @BSFishy , do you have any thoughts or suggestions on this?

BSFishy:
> While investigating this issue, I found it challenging to identify the bug due to the implicit typing of variables.
> 
> For example, the main issue lies in the fact that `palette[color]` is already a `string` in the format "#017D73". Therefore, the original code `rgbToHex(palette[color].rgba).toUpperCase()` is unnecessary and prone to errors.
> 
> Do we have any plans to migrate some of these files to TypeScript for better type safety?

Unfortunately, not really. These variables are extracted from Sass, using custom scripts: https://github.com/opensearch-project/oui/blob/main/scripts/babel/variables-from-scss/index.js and https://github.com/opensearch-project/oui/blob/main/scripts/lib/compile-scss-with-variables.js.

I think the format changed from

```ts
color: {
  r: string;
  g: string;
  b: string;
  a: string;
  rgba: string;
}
```

into

```ts
color: string
```

Which is where the errors come from. We might be able to add some sort of type safety by updating the Webpack plugin.

> Why do we repeat the colour themes using variables that have the suffix `Text`. For example, take a look at the screenshot below. We are repeating the values for hex colour`"#F66"` in the keys`euiDangerColor` and `euiDangerColorText`. In both cases, the value is a string. Shouldn't we avoid DRY?

To this point, the generates vars are just all of the Sass vars in the projects put into a JSON object. So for whatever reason we have these variables which are used in different ways, all just extracted and put into a single object.

BigSamu:
> To this point, the generates vars are just all of the Sass vars in the projects put into a JSON object. So for whatever reason we have these variables which are used in different ways, all just extracted and put into a single object.

Just asking here if we should simplify the JSON object to avoid DRY. It does not make sense to have the same value for 2 distinct keys. Of course, this will mean refactoring the code in other parts.

Regards,

Samuel

joshuarrrr:
> While investigating this issue, I found it challenging to identify the bug due to the implicit typing of variables.
> 
> For example, the main issue lies in the fact that `palette[color]` is already a `string` in the format "#017D73". Therefore, the original code `rgbToHex(palette[color].rgba).toUpperCase()` is unnecessary and prone to errors.
> 
> Do we have any plans to migrate some of these files to TypeScript for better type safety?

I don't think we have any overall issue for migrating OUI from JS to TS, but I think you could open one with this as an example of why it may be worth the effort to do. The good thing about typescript conversions is that they can be parallelized and worked on as good first issues.

joshuarrrr:
> > To this point, the generates vars are just all of the Sass vars in the projects put into a JSON object. So for whatever reason we have these variables which are used in different ways, all just extracted and put into a single object.
> 
> Just asking here if we should simplify the JSON object to avoid DRY. It does not make sense to have the same value for 2 distinct keys. Of course, this will mean refactoring the code in other parts.
> 
> Regards,
> 
> Samuel

@BigSamu This has to do with the way we use SASS vars in the project. In many cases, it's useful to have multiple aliases that map to different usages even if (in some themes) they actually use the same underlying value. When you look at the JSON file (or the output of `_get_sass_vars.js`), what you're seeing is the calculated/resolved values for each SASS var _in the currently selected theme_. But different themes (or even light/dark mode within a particular theme) may assign these values differently.

In your example, in that theme the `euiColorDanger` is suitable for use as a text color given the themed background colors, so it's not transformed by the `makeHighContrastColor()` function: https://github.com/search?q=repo%3Aopensearch-project%2Foui+ouiColorDangerText%3A&type=code

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@
 
 - Add exit code to compile-scss script on failure ([#1024](https://github.com/opensearch-project/oui/pull/1024))
 - Correct file path for import of Query component ([#1069](https://github.com/opensearch-project/oui/pull/1069))
+- Resolve "Guidelines" Documentation Render Blank Page ([#1111](https://github.com/opensearch-project/oui/pull/1111))
```
joshuarrrr, line 21:
```suggestion
- Fix "Guidelines" documentation links rendering blank pages ([#1111](https://github.com/opensearch-project/oui/pull/1111))
```

src-docs/src/views/guidelines/sass.js
```diff
@@ -113,7 +112,7 @@ function renderPaletteColor(palette, color) {
       <OuiFlexItem grow={false}>
         <div
           className="guideSass__swatch"
-          style={{ background: rgbToHex(palette[color].rgba).toUpperCase() }}
+          style={{ background: palette[color] }}
```
joshuarrrr, line 115: @Willie-The-Lord Is `palette[color]` already uppercased, or should we keep `.toUpperCase()` for consistency?

Willie-The-Lord:
> > > To this point, the generates vars are just all of the Sass vars in the projects put into a JSON object. So for whatever reason we have these variables which are used in different ways, all just extracted and put into a single object.
> > 
> > 
> > Just asking here if we should simplify the JSON object to avoid DRY. It does not make sense to have the same value for 2 distinct keys. Of course, this will mean refactoring the code in other parts.
> > Regards,
> > Samuel
> 
> @BigSamu This has to do with the way we use SASS vars in the project. In many cases, it's useful to have multiple aliases that map to different usages even if (in some themes) they actually use the same underlying value. When you look at the JSON file (or the output of `_get_sass_vars.js`), what you're seeing is the calculated/resolved values for each SASS var _in the currently selected theme_. But different themes (or even light/dark mode within a particular theme) may assign these values differently.
> 
> In your example, in that theme the `euiColorDanger` is suitable for use as a text color given the themed background colors, so it's not transformed by the `makeHighContrastColor()` function: https://github.com/search?q=repo%3Aopensearch-project%2Foui+ouiColorDangerText%3A&type=code
> 
> But each theme is free to define the relationship between the `euiColorDanger` and `euiColorDangerText` variables (even hard-coding both), and OUI components are structured to use the appropriate variable whether the same or different.

Thanks @joshuarrrr  for the detailed explanation! It makes sense that the flexibility is beneficial, especially when we consider the implementation across various themes.

src-docs/src/views/guidelines/colors/vis_palette.js
joshuarrrr: Currently visualization colors are intentionally defined to be light/dark mode agnostic, but we do have plans to make them different between themes:  https://github.com/opensearch-project/oui/issues/818.

So I think we can remove the TODO from the comment for now:
```suggestion
  // ouiPaletteColorBlind() is currently shared across themes and dark/light modes. This will need to be  made dynamic once we have different visualization palettes as in https://github.com/opensearch-project/oui/issues/818
```

joshuarrrr (commented):
Awesome fix @Willie-The-Lord - just a couple minor questions and suggestions.

src-docs/src/views/guidelines/sass.js
Willie-The-Lord: It's already uppercased.

Willie-The-Lord: The hex color code originates from this location and is already in uppercase: https://github.com/Willie-The-Lord/oui/blob/issue1109/src/global_styling/variables/_colors.scss#L16 Should we still include `.toUpperCase()` as a precaution, in case the file changes in the future and potentially breaks something?

joshuarrrr: nah, that's fine - thanks for the explanation

joshuarrrr (approved):

BigSamu:
> > > To this point, the generates vars are just all of the Sass vars in the projects put into a JSON object. So for whatever reason we have these variables which are used in different ways, all just extracted and put into a single object.
> > 
> > 
> > Just asking here if we should simplify the JSON object to avoid DRY. It does not make sense to have the same value for 2 distinct keys. Of course, this will mean refactoring the code in other parts.
> > Regards,
> > Samuel
> 
> @BigSamu This has to do with the way we use SASS vars in the project. In many cases, it's useful to have multiple aliases that map to different usages even if (in some themes) they actually use the same underlying value. When you look at the JSON file (or the output of `_get_sass_vars.js`), what you're seeing is the calculated/resolved values for each SASS var _in the currently selected theme_. But different themes (or even light/dark mode within a particular theme) may assign these values differently.
> 
> In your example, in that theme the `euiColorDanger` is suitable for use as a text color given the themed background colors, so it's not transformed by the `makeHighContrastColor()` function: https://github.com/search?q=repo%3Aopensearch-project%2Foui+ouiColorDangerText%3A&type=code
> 
> But each theme is free to define the relationship between the `euiColorDanger` and `euiColorDangerText` variables (even hard-coding both), and OUI components are structured to use the appropriate variable whether the same or different.

Many thanks @joshuarrrr! Pretty clear the explanation!

PR #1113, opened 2023-10-24
[Fix]  Rename crossInACircleFilled Asset Files and Ensure Backward Compatibility

### LINKED ISSUE #1045: Rename `crossInACircleFilled` to `crossInCircleFilled`
https://opensearch.slack.com/archives/C05389T9LJC/p1695072084427309?thread_ts=1694535707.701519&cid=C05389T9LJC

The files should be renamed to not have the `A` and the icon type should be `crossInCircleFilled` but `crossInACircleFilled` should still point to `crossInCircleFilled` for backwards compatibility

Willie-The-Lord:
### Description

- Renames asset files from `crossInACircleFilled` to `crossInCircleFilled`.
- Maps existing references to the new asset name.
- Implements a helper function to normalize both old and new names for backward compatibility.
- https://opensearch.slack.com/archives/C05389T9LJC/p1695072084427309?thread_ts=1694535707.701519&cid=C05389T9LJC

### Issues Resolved
closes #1045 

### Screenshot
Screenshot 2023-11-01 at 12 09 51 AM

### Check List
- [x] All tests pass
  - [x] `yarn lint`
  - [x] `yarn test-unit`


CMDWillYang (approved):
LGTM!

ss-vijayvavdiya (approved):
LGTM!

vvavdiya (approved):
LGTM

BigSamu (approved):
LGTM!

joshuarrrr:
@Willie-The-Lord Sorry, the original issue was a little unclear. But we need to still support usage of `crossInACircleFilled` as the icon name for backwards compatibility (see https://github.com/search?q=org%3Aopensearch-project%20crossInACircleFilled&type=code for some usage). The goal is to have to have the documentation website list the name as `crossInCircleFilled`, and to use that as the canonical name for assets, etc, but to also provide an alias between the old and new name, so that any existing usages of `crossInACircleFilled` correctly use the `crossInCircleFilled` assets.

Willie-The-Lord:
@joshuarrrr, to confirm my understanding and clarify the objectives:

- Rename the asset files to `crossInCircleFilled`.
- Map the existing `crossInACircleFilled` references to the newly named asset files.
- Update the documentation to reflect the name change to `crossInCircleFilled`.

My solution here is:

- Keep the existing code that references `crossInACircleFilled` untouched.
- Implement a helper function to normalize both names, thereby ensuring backward compatibility.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@
 - Update ouiTextSubduedColor in `next` dark theme ([#973](https://github.com/opensearch-project/oui/pull/973))
 - Add `crossInCircleEmpty` and `power` icons ([#1044](https://github.com/opensearch-project/oui/pull/1044))
 - Add `Figma` icon and link to OUI Figma resources ([#1064](https://github.com/opensearch-project/oui/pull/1064))
+- Rename `crossInACircleFilled` Asset Files and Ensure Backward Compatibility ([#1113](https://github.com/opensearch-project/oui/pull/1113))
```
joshuarrrr, line 16: Prefer sentence case for changelog entries

```suggestion
- Rename `crossInACircleFilled` asset files and ensure backward compatibility ([#1113](https://github.com/opensearch-project/oui/pull/1113))
```

src-docs/src/views/icon/icons.js
```diff
@@ -279,6 +279,14 @@ export const iconTypes = [
...
 ];
 
+// Add mapping icon name for backwards compatibility
+const normalizeIconType = (iconType) => {
+  const mapping = {
+    crossInACircleFilled: 'crossInCircleFilled',
+  };
+  return mapping[iconType] || iconType;
+};
+
```
joshuarrrr, line 289: @Willie-The-Lord Revert this function and its usage, as it's not needed here. What you want to do instead is replace `crossInACircleFilled`with `crossInCircleFilled` in `iconTypes` at L74.

It's important to understand how this `iconList` is used and why it's different from the `typeToPathMap` in `src/components/icon`. The list in this file is icons that will appear in the documentation website in the icon list: https://oui.opensearch.org/1.3/#/display/icons

So, because we only want the "correct" spelling of the icon to appear in the docs, we don't want to include the mispelling in that list. Note that this list _also_ is used by the doc-site-only "iconValidator" function: https://github.com/opensearch-project/oui/blob/main/src-docs/src/services/playground/iconValidator.js

By leaving the misspelled version off this list, when users try it in a playground component on the docs site, the validator will block it as an undocumented icon type, which is actually a nice subtle hint. However, if you actually use the old misspelled version as an icon type, it will work anywhere outside the doc site (because we don't use that validator, we use the actual types from the mapping. To see what I mean, after changing this file, you can try it with the `iconType` property in the playground tab of http://localhost:8030/#/display/badge/playground.

Basically, the list in this file is not all possible iconType strings, but the list that we encourage developers to use.

src/components/icon/__snapshots__/icon.test.tsx.snap
```diff
@@ -1656,6 +1656,23 @@ exports[`OuiIcon props type crossInCircleEmpty is rendered 1`] = `
 </svg>
 `;
 
+exports[`OuiIcon props type crossInCircleFilled is rendered 1`] = `
```
joshuarrrr, line 1659: nice - notice that the snapshot still exists for `crossInACircleFilled`, which is what we want for backwards compatibility.

src/components/icon/icon.tsx
```diff
@@ -111,7 +111,8 @@ const typeToPathMap = {
   crossClusterReplicationApp: 'app_cross_cluster_replication',
   crosshairs: 'crosshairs',
   crossInCircleEmpty: 'crossInCircleEmpty',
-  crossInACircleFilled: 'crossInACircleFilled',
+  crossInACircleFilled: 'crossInCircleFilled',
+  crossInCircleFilled: 'crossInCircleFilled',
```
joshuarrrr, line 115: nit - keys should be in alphabetical order. This mapping is the important bit that tells us which asset to use for a given type.

src/components/icon/assets/crossInCircleFilled.js
joshuarrrr, line 1: Please also update the const name in this file as well.

joshuarrrr (changes requested):
@Willie-The-Lord This is close, but there are a couple things left to change. Let me know if you still have any questions about how these various pieces work together.

src-docs/src/views/icon/icons.js
Willie-The-Lord: > However, if you actually use the old misspelled version as an icon type, it will work anywhere outside the doc site (because we don't use that validator, we use the actual types from the mapping.)

Ahh! That totally makes sense! I kinda mixed up the concept of the `iconType` and its usage. Now everything is making sense! Thanks for the clarification. By the way, the playground tip actually makes everything even clearer. Thanks, @joshuarrrr.

src/components/icon/icon.tsx
Willie-The-Lord: Isn't the key already ordered in alphabetical order?

src/components/icon/__snapshots__/icon.test.tsx.snap
Willie-The-Lord: I see what's going on right now. Much appreciated!

src/components/icon/icon.tsx
joshuarrrr: I think it should be crossInACircleFilled, crossInCircleEmpty, crossInCircleFilled - it's L113 that needs to go in between.

joshuarrrr (approved):
@Willie-The-Lord Thanks for the persistence and patience on this - looks great!

Willie-The-Lord:
> @Willie-The-Lord Thanks for the persistence and patience on this - looks great!

Np. I'm really enjoying the process~

BSFishy (approved):

PR #1119, opened 2023-10-28
[OSCI][FIX] responsive basic table components

### LINKED ISSUE #967: [BUG] OuiBasicTable Action icons within actions column becomes obscured @ small size
**Describe the bug**

At a small size, action icons within the actions column become obscured by the column contents to the left. 

**To Reproduce**
1. Go to [OuiBasicTable documentation](https://oui.opensearch.org/1.2/#/tabular-content/tables#adding-actions-to-table)
2. Toggle "multiple actions" on
3. Resize browser window 
4. Mouse over actions column to see example represented in screenshot

**Expected behavior**
The actions column is never obscured by another columns content. I would expect that if "multiple actions" is toggled on within any OuiBasicTable implementation within a product, that the responsive table switch would happen at a different target size - OR - the Actions within the column would only show the "all actions" button at this size. 

**Screenshots**

Screen Shot 2023-08-15 at 5 34 01 PM

thanhinhchtom:
### Description
Fix responsive basic table components

### Issues Resolved
Fix #967 

### Check List
- [x] All tests pass
  - [x] `yarn lint`
  - [x] `yarn test-unit`

### DIFF
--- a/src/components/basic_table/basic_table.tsx
+++ b/src/components/basic_table/basic_table.tsx
@@ -44,9 +44,12 @@ import {
   formatDate,
   formatNumber,
   formatText,
+  getBreakpoint,
   LEFT_ALIGNMENT,
+  OuiBreakpointSize,
   RIGHT_ALIGNMENT,
   SortDirection,
+  throttle,
 } from '../../services';
 import { CommonProps } from '../common';
 import { isFunction } from '../../services/predicate';
@@ -310,6 +313,7 @@ export type OuiBasicTableProps<T> = CommonProps &
 interface State<T> {
   initialSelectionRendered: boolean;
   selection: T[];
+  currentBreakpoint: OuiBreakpointSize | undefined;
 }
 
 interface SortOptions {
@@ -326,7 +330,7 @@ function hasPagination<T>(
   return x.hasOwnProperty('pagination') && !!x.pagination;
 }
 
-export class OuiBasicTable<T = any> extends Component<
+export class OuiBasicTable<T extends {} = any> extends Component<
   OuiBasicTableProps<T>,
   State<T>
 > {
@@ -375,12 +379,16 @@ export class OuiBasicTable<T = any> extends Component<
       // used for checking if  initial selection is rendered
       initialSelectionRendered: false,
       selection: [],
+      currentBreakpoint: getBreakpoint(
+        typeof window === 'undefined' ? -Infinity : window.innerWidth
+      ),
     };
   }
 
   componentDidMount() {
     if (this.props.loading && this.tbody) this.addLoadingListeners(this.tbody);
     this.getInitialSelection();
+    window.addEventListener('resize', this.functionToCallOnWindowResize);
   }
 
   componentDidUpdate(prevProps: OuiBasicTableProps<T>) {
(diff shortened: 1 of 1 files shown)


joshuarrrr:
@thanhinhchtom For UI changes like this, can you provide before/after screenshots or screencasts that demonstrate the behavior? That makes it easier for reviewers to understand the intended change and how it was tested.

thanhinhchtom:
@joshuarrrr I already added some description of what I have done to fix the responsive and visualize it with 2 screenshots above. Is that okay?

joshuarrrr:
> @joshuarrrr I already added some description of what I have done to fix the responsive and visualize it with 2 screenshots above. Is that okay?

Yep, thanks, that's helpful!

thanhinhchtom:
@joshuarrrr I already try to make it show 1 button in this situation, but I think the component that is currently used will make it always shows up even if I hover or not when I reduce it to 1 button. So I decided to not show anything as the recommended solution in the issue. Is it okay to do so?

src/components/basic_table/basic_table.tsx
```diff
@@ -375,12 +379,16 @@ export class OuiBasicTable<T = any> extends Component<
       // used for checking if  initial selection is rendered
       initialSelectionRendered: false,
       selection: [],
+      currentBreakpoint: getBreakpoint(
+        typeof window === 'undefined' ? -Infinity : window.innerWidth
+      ),
```
joshuarrrr, line 384: @thanhinhchtom I like the idea here, but I don't actually think this behavior should be dependent on the breakpoint, which is defined by the width of the browser window. Instead, what really matters is the width of the action column - we should only render the action icons on hover if there's room for them to render next to the three-dot-menu icon. 

There is one complication, though, which is the fact that the table becomes responsive at small breakpoints, and the icons no longer appear on hover.

So to summarize, rather than set up a listener on the window size, you should just check the width of the action column (assuming breakpoint > s). And render however many icons (0, 1, or 2) will fit within that column width. I think you still may need a listener for cases where the overall table width changes, because column size can be set as a percentage of the table.

joshuarrrr (changes requested):
@thanhinhchtom Thanks for the contribution - I think the code is mostly correct for what it does, but it's not quite the solution we'll want here.

src/components/basic_table/basic_table.tsx
thanhinhchtom: @joshuarrrr So I should not use breakpoint here. Instead I should use size of the columnand just fit more button if it is still enough space, right?

joshuarrrr: yes, that's correct. Although you may also still want to have your logic skip small breakpoints, because the different layout there already handles this problem.

thanhinhchtom: Okay, I will try that and ask you later if needed?

thanhinhchtom: @joshuarrrr One more quick question. I want to clarify the requirement. So I can still use the breakpoint there if needed, but I just cannot hardcode for the different situation as what I did and I have to find the way to use the current size of the component to decide how many buttons will show up, right?

joshuarrrr: The breakpoint usage should just be to determine whether column widths matter at all or not. But yeah the conditional logic for what and how to render should depend on the action column width. And the action column width may be itself dependent on the table/component width.

joshuarrrr:
Wetting this to draft - @thanhinhchtom Feel free to set it back if you're able to revisit.

PR #164, opened 2022-12-23
[OUI] Removed figma link/icon, Codesandbox link/icon, Sketch link/icon Casc…

AbhishekReddy1127:
Signed-off-by: AbhishekReddy1127 <nallamsa@amazon.com>

### Description
Removed Figma, Codesandbox, Sketch link and icon.
Removed Cascadia options and Babelfish from theme dropdown.
 
### Issues Resolved
#111 
 
### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [x] All tests pass
  - [x] `yarn lint`
  - [x] `yarn test-unit`
- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/oui/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1152, opened 2023-11-14
[Security] Remediate `axios` vulnerabilities and remove code sandbox links

joshuarrrr:
### Description

Remove CodeSandbox links from Demo JS tabs as follow-up to https://github.com/opensearch-project/oui/pull/164 (but not the codesandbox icon, which would be breaking)

Add resolution for `wait-on` (I'd prefer to bump `start-server-and-test`, but we can't because they [dropped support for node 14](https://github.com/bahmutov/start-server-and-test/releases/tag/v2.0.1)).

I made the decision that maintaining the code sandbox integration in it's current form was not longer worth it, due to the fact that https://www.npmjs.com/package/codesandbox hasn't been updated in 3 years and we were adding an increasing number of resolutions to support it, even though it doesn't currently work (due to the way we handled the `react-datepicker` package). But if folks think it's a useful or worthwhile feature to fix and restore, we may want to just start again from scratch.

Current code sandbox link usage:

https://github.com/opensearch-project/oui/assets/1679762/37aab22c-82df-473b-8eb0-0510006cc6ab

### Issues Resolved

Fixes https://github.com/opensearch-project/oui/issues/1149, https://github.com/opensearch-project/oui/issues/1148

### Check List
- [ ] All tests pass
  - [ ] `yarn lint`
  - [ ] `yarn test-unit`


src-docs/src/components/guide_section/guide_section_parts/guide_section_code.tsx
```diff
@@ -1,54 +0,0 @@
...
-
-  const codeSandboxLink = (
-    <CodeSandboxLink
-      className="guideSectionExampleCode__link"
-      content={code.default}>
-      <OuiButtonEmpty size="xs" iconType="logoCodesandbox">
-        Try out this demo on Code Sandbox
-      </OuiButtonEmpty>
-    </CodeSandboxLink>
-  );
```
BSFishy, line 44: This PR breaks demo JS sections on the docsite. I think this file can be kept just without this section

joshuarrrr: Interesting - I remember searching for any usage of codeSandBox links/blocks and coming up empty. Do you happen to have some example URLs on the doc site I can use for reference?

BSFishy: The issue is that the Demo JS section errors out with this change:

Screenshot (193)

If I'm not mistaken, it's because this file is removed

joshuarrrr: D'oh! Thanks, I'll fix. I was a little too aggressive with my deleting 🌪️

joshuarrrr: Fixed in latest revision

src-docs/src/views/form_controls/display_toggles.js
```diff
@@ -199,7 +199,6 @@ DisplayToggles.propTypes = {
   canAppend: PropTypes.bool,
   canInvalid: PropTypes.bool,
   extras: PropTypes.arrayOf(PropTypes.node),
-  // Manually building the spacer array to avoid having to import Spacer into codesandbox
```
BSFishy, line 202: So if this was done manually because of codesandbox and we're no longer supporting codesandbox, couldn't this be generated automatically?

joshuarrrr: Yes...? 😆 

I think my thought process was along the lines of, what if the comment was only the _original_ reason it was done this way, but over time other things relied on that behavior without updating the comment.

But I'll try updating it to be more DRY.

AMoo-Miki (approved):

ashwin-pc (approved):

Issue #1129, opened 2023-05-18
Hover over the filter is not showing text in vertical bar visualization in OpenSearch dashboards


tejashu7:
Description:
Hover over the filter is not showing text in vertical bar visualization in OpenSearch dashboars.

1. Create a vertical dashboard with 2 legends.
2. Click on the legend it will open the color palette and filter options.
3. Hover on the “+” → filter out value which appears after the legend is clicked for color palette filter as shown for Kibana. The hover text over that “+/-” sign is not appearing for OpenSearch-dashboards.

**Expected behavior**
A text descipbing the filter out value should be shown when hovered over the "+" sign.

**OpenSearch Version**
 OpenSearch 2.6.0

**Dashboards Version**
OpenSearch Dashboard version is 2.6.0

**Plugins**

Please list all plugins currently enabled.

**Screenshots**

**Host/Environment (please complete the following information):**
 - Browser and version [e.g. 22]

**Additional context**

Please find the forum link for further details:
https://forum.opensearch.org/t/hover-over-filter-in-vertical-bar-in-opensearch-dashboards-not-showing-text/14124

BigSamu:
@joshuarrrr,

May I take a look at this issue? Feeling is something straightforward or not?

Regards,

Samuel

joshuarrrr:
Yep, @BigSamu , I think this would be a good issue for you to do a little more investigation on. We'll still need Kevin to provide some guidance on the intended behavior, but you can dig into the implementation of the `isIconOnly` prop to see how we might best fix it.

kgcreative:
@joshuarrrr + @BigSamu -- For accessibility (and usability) reasons, if the form element doesn't have a label (which looks like button groups probably wouldn't), then a tooltip (https://oui.opensearch.org/1.3/#/display/tooltip) should display that information. `filter for` and `filter out` make sense here, although a nice to have would be `filter for {key:value}`

BigSamu:
> Yep, @BigSamu , I think this would be a good issue for you to do a little more investigation on. We'll still need Kevin to provide some guidance on the intended behavior, but you can dig into the implementation of the `isIconOnly` prop to see how we might best fix it.

Great! Thanks @joshuarrrr !

[...]

PR #1160, opened 2023-11-17
[OSCI][BUG] Fixing bug in icon-only buttons for showing info tooltip when hovered

BigSamu:
### Description
Icon-only buttons are not showing descriptive tooltip when hovered. This PR fixes the problem by adding wrapping a <span> tag with `textProps` to a `<buttonIcon/>` component

### Issues Resolved
Issue #1129 

### Check List
- [x] All tests pass
  - [x] `yarn lint`
  - [x] `yarn test-unit`


src/components/button/button_content.tsx
```diff
@@ -109,7 +109,9 @@ export const OuiButtonContent: FunctionComponent<
 
   return (
     <span {...contentProps} className={contentClassNames}>
-      {buttonIcon}
+      <span {...textProps} className="">
```
BSFishy, line 112:
```suggestion
      <span {...textProps}>
```

BigSamu: Hi @BSFishy,

Thanks for the comment. Regarding your suggestion of removing `className=""` I already explained @joshuarrrr and @kgcreative the reason behind this in [this](https://github.com/opensearch-project/oui/issues/1129#issuecomment-1804846572) comment. Here that explanation again below:

>The dictionary textProps for the case of iconOnly button contains the following:

>{className: 'ouiButton__text ouiButtonGroupButton__textShift', data-text: 'Option three', title: 'Option three', ref: ƒ}

>By adding, className="" after {...textProps} we override the initial definition of className coming in textProps. This simple solution works allowing us to visualize the tooltip. Of course, I feel it is not quite elegant, and maybe we should move the HTML attributes data-text and title to the contentProps

I can create some custom styling to create a more ad-hoc solution if you want

BSFishy: Okay I see. In that case, I think it would be better to copy `textProps` and extract `className` out of it. Adding `className` like this is a clever solution, but it isn't clear just by code what it's doing. Additionally, it muddies up the snapshots with empty classnames.

Maybe something like

```tsx
const iconTextProps = { ...textProps };
iconTextProps.className = undefined;
```

BigSamu: Yeah! I new my solution was not elegant. Thanks for the suggestion @BSFishy. I will implement it and let you know when I the update is ready

BSFishy: Thinking about it more, there are potentially a few other options

Something like

```tsx
const { className: _, ...iconTextProps } = textProps;
```

or doing

```tsx
<span {...textProps} className={undefined}>
```

I think should also work. The only reason I bring it up is because my first suggestion might unnecessarily create memory, but I don't know do whatever sounds good to you

BigSamu: I went with the second option. Code updated. Let me know if you have any concerns.

BSFishy:
> Is there any issue with this? Do you know what could be causing this?

This seems to be a bigger issue. I did a little bit of digging and found that the docinfo isn't being generated for `OuiButton` for some reason. I'll make an issue for this.

src/components/accessibility/__snapshots__/skip_link.test.tsx.snap
```diff
@@ -11,6 +11,7 @@ exports[`OuiSkipLink is rendered 1`] = `
   <span
     class="ouiButtonContent ouiButton__content"
   >
+    <span />
```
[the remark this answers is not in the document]
BigSamu: Makes total sense. It's a bit new when using snapshoots for testing, but now I see their usefulness in this case. Yeaph! I will include a logic for a conditional rendering.

src/components/button/button_content.tsx
BigSamu, line 1: @joshuarrrr,

As requested, I made the changes you suggested. Now, I am passing the title prop inside the definition of the `buttonIcon` variable (an `OuiIcon` component). Now the fix makes more clear the code and avoid creating empty spans in cases where are not button icons

src/components/button/button_content.tsx
```diff
@@ -91,11 +91,13 @@ export const OuiButtonContent: FunctionComponent<
...
+    const { title } = { ...textProps };
     buttonIcon = (
       <OuiIcon
         className="ouiButtonContent__icon"
         type={iconType}
         size={iconSize}
+        title={title}
         color="inherit" // forces the icon to inherit its parent color
       />
     );
```
BigSamu, line 103: @joshuarrrr,

As requested, I made the changes you suggested. Now, I am passing the title prop inside the definition of the `buttonIcon` variable (an `OuiIcon` component). Now the fix makes more clear the code and avoid creating empty spans in cases where are not button icons

src/components/button/button_content.tsx
```diff
@@ -109,9 +111,7 @@ export const OuiButtonContent: FunctionComponent<
 
   return (
     <span {...contentProps} className={contentClassNames}>
-      <span {...textProps} className={undefined}>
-        {buttonIcon}
-      </span>
+      {buttonIcon}
       <span {...textProps}>{children}</span>
```
BigSamu, line 115: As mentioned before, no the wrapping `<span>` element is removed from `buttonIcon` element

BigSamu:
@BSFishy, @joshuarrrr,
 
Last updates requested are done. I added comments in a code review above so you can check the exact changes I made. I am still a bit confused about how to use snapshots. But I would bring that question to the following community office hours.
 
 Regards,
 
 Samuel

src/components/button/button_content.tsx
```diff
@@ -91,11 +91,13 @@ export const OuiButtonContent: FunctionComponent<
       <OuiLoadingSpinner className="ouiButtonContent__spinner" size="m" />
     );
   } else if (iconType) {
+    const { title } = { ...textProps };
     buttonIcon = (
       <OuiIcon
         className="ouiButtonContent__icon"
         type={iconType}
         size={iconSize}
+        title={title}
```
BSFishy, line 100: I feel like we could just simplify this whole thing to

```suggestion
        title={textProps.title}
```

and delete line 94, but I won't block on that

BSFishy (approved):

src/components/button/button_content.tsx
BigSamu: Good suggestion. But it could be the case that textProps is undefined. Thus I include an optional changing operator to what you suggest as follows:

``js
title={textProps?.title}
``

ashwin-pc (approved):

ashwin-pc (approved):

src/components/accessibility/__snapshots__/skip_link.test.tsx.snap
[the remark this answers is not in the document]
BigSamu: Resolved!

BigSamu:
@joshuarrrr,

Revies asked are resolved. Could you check for proceeding with a merge?

joshuarrrr (approved):

PR #1164, opened 2023-11-21
[Fix] Theme selector incorrectly focus on first item after selection

### LINKED ISSUE #1163: [BUG] Theme selector incorrectly focus on first item after selection
**Describe the bug**

When the theme selector is opened, the focus incorrectly remains on the first menu item.

**To Reproduce**

1. Click the theme selection popup button
2. Select any theme other than the first one from the list.

**Expected behavior**

I think the checkmark icon solely is enough, no need another indicator to highlight the current selection.  @opensearch-project/opensearch-ux please help confirm the correct UX~

**Screenshots**

https://github.com/opensearch-project/oui/assets/65143821/55851b51-e90c-4570-923c-032cfa1b760f

willie-hung:
### Description

This PR removes the initial focusing state when opening up the theme selector popup.

### Issues Resolved
Fixes #1163 

#### Before

https://github.com/opensearch-project/oui/assets/65143821/c15db0c1-fd19-48d9-a458-cab4642d459d

#### After

https://github.com/opensearch-project/oui/assets/65143821/6326c9d9-db29-45a0-87ef-fbcc2703195d

### Misc

Also fix a typo in `guide_version_selector.tsx`.

### Check List
- [ ] All tests pass
  - [ ] `yarn lint`
  - [ ] `yarn test-unit`

### DIFF
--- a/src-docs/src/components/guide_version_selector/guide_version_selector.tsx
+++ b/src-docs/src/components/guide_version_selector/guide_version_selector.tsx
@@ -78,7 +78,7 @@ export const GuideVersionSelector: FunctionComponent<GuideVersionSelectorProps> 
           href={`/${option}`}
           onClick={onChange(option)}>
           <OuiFlexGroup direction="row" wrap={false}>
-            <OuiFlexItem>v{option}</OuiFlexItem>
+            <OuiFlexItem>{option}</OuiFlexItem>
             {i === 0 && (
               <OuiFlexItem>
                 <OuiBadge>Latest</OuiBadge>
--- a/src/components/basic_table/__snapshots__/collapsed_item_actions.test.tsx.snap
+++ b/src/components/basic_table/__snapshots__/collapsed_item_actions.test.tsx.snap
@@ -45,7 +45,7 @@ exports[`CollapsedItemActions render with href and _target provided 1`] = `
   hasArrow={true}
   id="id-actions"
   isOpen={true}
-  ownFocus={true}
+  ownFocus={false}
   panelPaddingSize="none"
   popoverRef={[Function]}
 >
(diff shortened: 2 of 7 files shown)


Willie-The-Lord:
Hi @kgcreative, could you please assist in determining the best UX for this scenario? Thank you!

kgcreative:
This is a challenging one. When opening via mouse, we shouldn't have a focus selector, however, per WCAG accessibility standards, there is an expectation that when you open a control via keyboard interaction, that we move keyboard focus to the first item in the opened control. As such, if we can't differentiate between keyboard or mouse input, the existing behavior is the more correct one.

Willie-The-Lord:
Thanks for sharing your insights, @kgcreative! Based on what you've described, would you suggest implementing a mechanism where if the element is opened via keyboard (for example, pressing the Tab key), we highlight the selector, but if not, we skip the highlighting?

kgcreative:
> Thanks for sharing your insights, @kgcreative! Based on what you've described, would you suggest implementing a mechanism where if the element is opened via keyboard (for example, pressing the Tab key), we highlight the selector, but if not, we skip the highlighting?

This may be a larger issue. I would explore using [`:focus-visible`](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible) instead of `:focus`, since that does use browser heuristics to determine whether an element should have a visible focus indicator or not.

cc: @BSFishy + @joshuarrrr

BSFishy:
I feel like this is the wrong way to go about this. I think that changing the default value for focus would be breaking for UX. Additionally, `OuiContextMenuPanelDescriptor` offers `initialFocusedItemIndex`, which I think would be a better way to implement this. Just set that value to whatever value is currently selected. That way we don't even need to differentiate if the panel was opened by keyboard or mouse

kgcreative:
Yeah, this needs to be a component level change, not changing the fous behavior in the implementation

AMoo-Miki:
@willie-hung are you interested in taking this forward or would you rather have it closed?

willie-hung:
Hi @AMoo-Miki, I think it should be closed for now. However, if you decide to make changes at the component level, I would be happy to join the discussion and assist with the development.

AMoo-Miki:
Closing as recommended.

PR #1186, opened 2023-12-15
icon computation without mutable variables

### LINKED ISSUE #1167: Rewrite initial icon computation without mutable variables
https://github.com/opensearch-project/oui/pull/1137#discussion_r1416035533

This seems like it could be extracted into a function and use return rather than a mutable variable.

ShatilKhan:
### Description
I have tried to use a helper function `getIconAndLoadingStatus` to return necessary values without having to mutate any variables.The returned values are then destructured into initialIcon and isLoading, which are used to set the state.

### Issues Resolved
Fixes #1167 

### Check List
- [ ] All tests pass
  - [ ] `yarn lint`
  - [ ] `yarn test-unit`


ShatilKhan:
@BSFishy 
Can you let me know if this is the right approach?

src/components/icon/icon.tsx
```diff
@@ -625,21 +625,27 @@ export class OuiIcon extends PureComponent<OuiIconProps, State> {
     let initialIcon;
     let isLoading = false;
```
joshuarrrr, line 626: These should be removed now, right?

src/components/icon/icon.tsx
```diff
@@ -625,21 +625,27 @@ export class OuiIcon extends PureComponent<OuiIconProps, State> {
...
 
-    // Category 1: cached oui icons
+    // Define helper function
+    const getIconAndLoadingStatus = (type: any) => {
```
joshuarrrr, line 629: I think we should already have a type for `type` that's more narrow?

joshuarrrr (commented):
@ShatilKhan This definitely looks like it's on the right track. You can probably take it further by moving the function outside the current scope, so you can write independent unit tests for it (one of the benefits of refactoring from mutable variables to a pure utility fn).

PR #1285, opened 2024-07-09
feat: remove Button translations and add inset shadow on active

virajsanghvi:
### Description
To avoid buttons physically moving, we've removed the translations on hover/focus and replaced it with an inset shadow (with existing removal of focus/active shadow). We will likely change/refine this behavior in the future, but this removes the weirdness of a physically moving button.

I changed the implementation of ouiSlightShadowActive because there were no active consumers in OUI or OpenSearch Dashboards.

The main downsides of this approach are when background is black and the button is white or a button icon without a border as these cases won't have a super visible change on active.

| Item |  Before (v7dark) | After (v7dark) | Before (v8dark) | After (v8dark) |
| ---- | ---- | ---- | ---- | ---- | 
| Button | button v7 before| button v7 after|button v8 before |button v8 after |
| Empty Button (no change) |v7 empty before | v7 empty after| | |
| Button Icon |v7 buttonicon before | v7 buttonicon after| | |
| Toggle Button |v7 toggle before | v7 toggle after| | |
| Button Group (no change) | v7 group before| v7 group after| | |
| Ghost Button | v7 ghost before| v7 ghost before| | |
| Split Button |v7 split before|v7 split after | | |

fyi: @lauralexis 

### Issues Resolved
N/A

### Check List
- [X] New functionality has been documented.
- [X] All tests pass
  - [X] `yarn lint`
  - [X] `yarn test-unit`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,7 +28,7 @@
 - [Next Theme] Set base font size to 18px ([#1221](https://github.com/opensearch-project/oui/pull/1221))
 - [Next Theme] Revert `font-weight` of OuiButton to normal from semi-bold ([#1222](https://github.com/opensearch-project/oui/pull/1222))
 - Convert shorthand palette colors to full 6-char hex ([#1262](https://github.com/opensearch-project/oui/pull/1262))
-- Remove Button translations and add inset shadow on active ([]())
+- Remove Button translations and add inset shadow on active ([#1285](https://github.com/opensearch-project/oui/pull))
 
 ### 🐛 Bug Fixes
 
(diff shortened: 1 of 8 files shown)


ashwin-pc:
I'm good with removing the movement, but what's the reason for removing the shadow?

virajsanghvi:
> I'm good with removing the movement, but what's the reason for removing the shadow?

Adding the inset shadow on active had the effect of disabling the hover outer shadow. This is actually what some other design systems do - the inset basically makes the button looked depressed and the loss of the outer shadow in that state also helps in seeing something changed for colors that make the inset harder to see.

ashwin-pc:
Oh my bad, I assumed that you removed it. That works. Thanks for the clarification

ashwin-pc (approved):

PR #1473, opened 2024-12-09
(security): update webpack-dev-server to address cves

d-rowe:
### Description
This addresses 11 CVEs associated with webpack-dev-server v3.

| Resolved CVE | Vulnerable Library |
|-----|--------------------|
| CVE-2024-37890 | ws-6.2.2 |
| CVE-2024-43796 | express-4.19.2 |
| CVE-2024-45590 | body-parser-1.20.2 |
| CVE-2024-21536 | http-proxy-middleware-0.19.1 |
| CVE-2024-47764 | cookie-0.6.0 |
| CVE-2024-29415 | ip-1.1.9 |
| CVE-2024-29180 | webpack-dev-middleware-3.7.2 |
| CVE-2024-52798 | path-to-regexp-0.1.7 |
| CVE-2024-45296 | path-to-regexp-0.1.7 |
| CVE-2024-43800 | serve-static-1.15.0 |
| CVE-2024-43799 | send-0.18.0 |

### Issues Resolved
fixes #389 

### Verification
I was able to diff the `dist` artifacts with and without these changes. The only change across all of the generated artifacts is in the charts_theme module.
```
/******/ 	// Load entry module and return exports
- /******/ 	return __webpack_require__(__webpack_require__.s = "./themes/charts/themes.ts");
+ /******/ 	return __webpack_require__(__webpack_require__.s = 0);
/******/ })
```

I've validated that the imported chart theme object is identical as is the type definition file.

I've also manually validated the docs pages.

### Check List
- [ ] New functionality includes testing.
- [ ] New functionality has been documented.
- [x] All tests pass
  - [x] `yarn lint`
  - [x] `yarn test-unit`
- [x] Update [CHANGELOG.md](./../CHANGELOG.md)
- [x] Commits are signed per the DCO using --signoff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/oui/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #1567, opened 2025-06-25
Fix dev server configuration after webpack-dev-server upgrade in 1.19

virajsanghvi:
### Description
https://github.com/opensearch-project/oui/pull/1473 upgraded versions of webpack-cli and webpack-dev-server. After these changes, runtime errors from ResizeObservable bubbled up to the dev server overlay causing a loop and error for certain components. This doesn't impact the production build, and this code is already out in 1.19. 

The MV upgrade of webpack-cli removed a WEBPACK_DEV_SERVER env variable, which enables us to not show runtime errors in the overlay (fixing the issue above), but also pointed out that the dev server config wasn't updated (which I did here). 

I had to drop deasync because it caused an infinite loop (unclear why this changed behavior). Other sync libraries to get ports had other issues (cves, unnecessary dependencies) and using async/await in the same file failed because eslint doesn't support async webpack config functions. I went with just wrapping this config with an async/await function in a separate file just for dev server as it seemed simplest.

Also, Query in search bar started causing an issue as its getting re-exported. Not completely clear why, but I think its because of webpack version mismatches, and I'd rather spend time upgrade webpack than fixing this, so just put in a work around.

### Issues Resolved
N/A

### Check List
- [ ] All tests pass
  - [ ] `yarn lint`
  - [ ] `yarn test-unit`

### DIFF
--- a/src-docs/webpack.config.js
+++ b/src-docs/webpack.config.js
@@ -16,9 +16,6 @@ const CircularDependencyPlugin = require('circular-dependency-plugin');
 const babelConfig = require('./.babelrc.js');
 // const ForkTsCheckerWebpackPlugin = require('fork-ts-checker-webpack-plugin');
 
-const getPort = require('get-port');
-const deasync = require('deasync');
-
 const { NODE_ENV, CI, WEBPACK_DEV_SERVER } = process.env;
 
 const isDevelopment = WEBPACK_DEV_SERVER === 'true' && CI == null;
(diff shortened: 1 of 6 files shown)


src-docs/webpack.config.js
```diff
@@ -41,155 +40,142 @@ function employCache(loaders) {
   return loaders;
 }
 
-const webpackConfig = {
-  mode: isProduction ? 'production' : 'development',
+const webpackConfig = async function () {
+  return {
+    mode: isProduction ? 'production' : 'development',
```
virajsanghvi, line 45: This is just indented - recommend looking at devServer config for changes

ruchidh (approved):

angle943 (approved):

PR #1630, opened 2025-08-07
Add support running dev server and tests in react 18

virajsanghvi:
### Description
This PR adds react 18, the ability to use dev server or run tests in react 18, and setup for react testing library. The main point of this is to add a mechanism that the library works in react 18 (unit tests in CI)

Specifically:
- Adds react 18 dependencies under an alias so we can swap it
- Adds a package.json script to start dev server under react 18 -> this appears to work fully
- Adds a package.json script to run unit tests under react 18
  - Added the unofficial react 18 enzyme adapter (doesn't handle events well) 
  - Added setup for react testing library
  - Migrated two tests to react testing library
  - This does not fully pass yet (as expected), but just helps with validation
  
Callouts:
- About 20% of test suites have a failing test, so trying to prioritize migrating those to react testing library first 
- Will add react 18 unit tests to CI once its passing
- Using ai to migrate tests, so will post PRs incrementally as it gets through stuff

### Issues Resolved
N/A

### Check List
- [ ] All tests pass
  - [ ] `yarn lint`
  - [ ] `yarn test-unit`


scripts/jest/setup/throw_on_console_error_react18.js
```diff
@@ -0,0 +1,29 @@
...
+    message.includes('findDOMNode is deprecated')
+  ) {
+    // Just log the warning without throwing
+    originalConsoleError('Warning suppressed for React 18 tests:', message);
+    return;
+  }
+
+  // For all other errors, throw as before
+  throw new Error(message);
+};
```
angle943, line 29: Just looking through the codebase for this, is it true that `findDOMNode` is only used in our test files?

- Are we planning on refactoring those tests to not use `findDOMNode`?
- Should we create an issue to track the removal of those as well as removing this error override once they have been removed?

package.json
```diff
@@ -139,13 +141,19 @@
     "@babel/preset-react": "^7.10.4",
     "@babel/preset-typescript": "^7.12.1",
     "@babel/template": "^7.10.4",
+    "@cfaester/enzyme-adapter-react-18": "^0.8.0",
```
angle943, line 144: i think this is a temp solution right? Should we create an issue to migrate away from enzyme as well as removing this?

PR #1807, opened 2026-07-28
Fix npm pack inclusion for lifecycle scripts

### LINKED ISSUE #1808: [BUG] 1.24.0 is uninstallable: published tarball is missing scripts/preinstall.js and scripts/postinstall.js
## Summary

`@opensearch-project/oui@1.24.0` cannot be installed at all. Its published tarball is missing the whole `scripts/` directory, but `package.json` still declares install hooks that live there:

```json
"preinstall": "node ./scripts/preinstall.js",
"postinstall": "node ./scripts/postinstall.js"
```

So every install dies immediately:

```
error .../node_modules/@elastic/eui: Command failed.
Command: node ./scripts/preinstall.js
Error: Cannot find module '.../node_modules/@elastic/eui/scripts/preinstall.js'
```

| Version | files under `scripts/` in the published tarball |
|---|---|
| 1.22.1 | 2 (`preinstall.js`, `postinstall.js`) |
| 1.23.0 | 2 |
| **1.24.0** | **0** ❌ |

**No source file changed.** `.npmignore` is byte-identical between the `1.23.0` and `1.24.0` tags. What changed is *how the package gets published*.

## What changed between 1.23.0 and 1.24.0

1.24.0 is the first release published by the new release workflow (`.github/workflows/release-drafter.yml`, rewritten on `1.x` by #1792, which backported #1753 / #1755 / #1766).

| | up to 1.23.0 | 1.24.0 |
|---|---|---|
| Did the release workflow publish to npm? | **No.** It only built `oui.tar.gz` and attached it to a *draft GitHub release*. Publishing to npm happened by some other (non-workflow) route. | **Yes** — `npm publish` in CI |
| npm version doing the packing | n/a in CI; whatever published 1.23.0 honoured the `.npmignore` pattern | **npm 11.x** (the workflow pins Node 24, overriding `.nvmrc`, because trusted publishing needs npm ≥ 11.5.1) |
| `scripts/` in the resulting tarball | 2 files ✅ | 0 files ❌ |

That's the entire gap: **the npm tarball started being produced by `npm publish`, and npm's packer does not honour the pattern `.npmignore` uses.**

## Why that breaks it

`.npmignore` keeps the two install hooks with a shell *extglob* negation:

```
# ignore everything in `scripts` except postinstall.js and preinstall.js
scripts/!(postinstall.js|preinstall.js)
```

npm's packer stopped honouring extglob negation in **npm 8.12.0**. It does not degrade gracefully — it prunes the *entire* directory. Bisected with `npm pack --dry-run` on a fixture containing `preinstall.js`, `postinstall.js`, and one other file:

| Packer | `scripts/` in tarball |
|---|---|
| npm 6.14.18, 7.24.2, 8.0.0, 8.3.2, 8.5.5, 8.11.0 | both hooks kept ✅ |
| **npm 8.12.0** and everything after (8.12.2 … 8.19.4, 9.9.4, 10.9.8, 11.6.4) | whole dir dropped ❌ |
| yarn 1.22.22 | both hooks kept ✅ |

So two conditions had to line up, and **neither alone breaks anything**:

1. `.npmignore` relies on extglob negation — a latent bug present for years.
2. The release started using `npm publish` (npm ≥ 8.12) — which exposed it.

The release infra work did not introduce the defect. It stopped hiding it.

<details>
<summary>Notes on the npm 8.12.0 regression (for the curious)</summary>

`npm-packlist` (5.1.0) and `ignore-walk` (5.0.1) are byte-identical between npm 8.11.0 and 8.12.0, so this is a change in npm's own code, not those dependencies. I did not pin the exact commit. It is *not* minimatch's `noext` option — `noext: true` makes the pattern match nothing, which would keep everything, the opposite of what happens.
</details>

## One thing this is *not*

The same workflow rewrite also deleted a `yarn pack --filename oui.tar.gz` step. **That removal is not the cause**, and it's an easy thing to misread.

`npm publish` does its own packing; it never consumed the `yarn pack` output. That tarball was only `tar`'d into `artifacts.tar.gz` and attached to a GitHub release — it never reached the npm registry.

Proof: on `main`, #1708 added `npm publish` while *leaving `yarn pack` in place*. Both coexisted in the same file:

```
43:          yarn pack --filename oui.tar.gz     ← still present
51:      - run: npm publish                       ← already added
```

A release cut at that commit would have been broken *with `yarn pack` still there*. The `yarn pack` removal (#1753 on `main`) is inert here.

Also not a factor: the `.nvmrc` bump `18.16.0` → `22.22.0`. The publish step overrides `.nvmrc` with Node 24 anyway, and Node 18 already ships npm 9.5.1, which is equally affected.

## Impact: `--ignore-scripts` is not a workaround

`scripts/postinstall.js` generates the OUI→EUI theme aliases, and `src/themes/eui/` is **not shipped** in the tarball — it is created at install time. Skipping scripts just trades the install failure for a broken SCSS build:

| Install | `src/themes/` contents |
|---|---|
| 1.23.0 (postinstall ran) | `charts, eui, eui-next, oui, oui-next, v9` |
| 1.24.0 with `--ignore-scripts` | `charts, oui, oui-next, v9` — **no `eui/`** |

OpenSearch Dashboards imports those aliases directly from its core stylesheets (all four `_globals_v{7,8}{light,dark}.scss`), e.g.:

```scss
@import "@elastic/eui/src/themes/eui/eui_colors_light";
@import "@elastic/eui/src/themes/eui/eui_globals";
```

So 1.24.0 is unusable by any route.

## To reproduce

```bash
mkdir /tmp/oui-124 && cd /tmp/oui-124
cat > package.json <<'EOF'
{
  "name": "oui-install-verify",
  "version": "1.0.0",
  "private": true,
  "dependencies": {
    "@elastic/eui": "npm:@opensearch-project/oui@1.24.0"
  }
}
EOF
yarn install     # fails: Cannot find module '.../scripts/preinstall.js'
                 # change to 1.23.0 and it installs cleanly
```

Or inspect the tarballs directly:

```bash
npm pack @opensearch-project/oui@1.24.0
tar -tzf opensearch-project-oui-1.24.0.tgz | grep '^package/scripts/'   # no output
npm pack @opensearch-project/oui@1.23.0
tar -tzf opensearch-project-oui-1.23.0.tgz | grep '^package/scripts/'   # two files
```

## Expected behavior

1.24.0 installs successfully, ships both install hooks, and generates `src/themes/eui/` at install time — regardless of which packer performs the release.

## Suggested fix

Fix the `.npmignore` pattern, not the publish tooling. Gitignore-style negation is honoured by every packer tested (npm 6 → 11, and yarn 1):

```
scripts/*
!scripts/preinstall.js
!scripts/postinstall.js
```

Reverting to a yarn-based publish would also unblock the release, but it forfeits trusted publishing and leaves the trap armed for whoever next runs `npm publish`.

Fixes proposed in #1809 (`main`) and #1810 (`1.x`). Since release tags are cut from `1.x`, **#1810 is the one needed to unblock** — and merging alone does not republish, so a **1.24.1 patch release from `1.x`** is required.

## Host/Environment
- OUI Version: 1.24.0 (1.23.0 and earlier unaffected)
- OSD Version (if applicable): affects any consumer; reproduced against OSD `main`
- OS: Linux
- Toolchains tested: node 22.22.3; npm 6.14.18, 7.24.2, 8.0.0–8.19.4, 9.9.4, 10.9.8, 11.6.4; yarn 1.22.22

ruanyl:
### Description

Updates `.npmignore` to use explicit ignore/unignore rules for `scripts/preinstall.js` and `scripts/postinstall.js`.

The previous extglob rule `scripts/!(postinstall.js|preinstall.js)` is not handled consistently by newer npm pack tooling. As a result, published tarballs can omit both lifecycle scripts while `package.json` still references them.

### Verification

Against unmodified `base/main`:

```
npm@8.19.4  postinstall=true  preinstall=true  scriptCount=2
npm@9.9.4   postinstall=false preinstall=false scriptCount=0
npm@11.18.0 postinstall=false preinstall=false scriptCount=0
```

Against this branch:

```
npm@8.19.4  postinstall=true preinstall=true scriptCount=2
npm@9.8.1   postinstall=true preinstall=true scriptCount=2
npm@9.9.4   postinstall=true preinstall=true scriptCount=2
npm@11.18.0 postinstall=true preinstall=true scriptCount=2
```

closed #1808

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -14,7 +14,9 @@ packages/eslint-plugin
 types/
 
 # ignore everything in `scripts` except postinstall.js and preinstall.js
-scripts/!(postinstall.js|preinstall.js)
+scripts/*
+!scripts/postinstall.js
+!scripts/preinstall.js
 
 src/**/*.!(scss)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@
 
 ### 🐛 Bug Fixes
 
+- Fix npm pack inclusion for lifecycle scripts with newer npm versions ([#1807](https://github.com/opensearch-project/oui/pull/1807))
+
 ### 🚞 Infrastructure
 
 ### 📝 Documentation


ruanyl:
closed, duplicate of https://github.com/opensearch-project/oui/pull/1809

